PR #46, opened 2021-03-23
Call/cc + Garbage collector fixes + Async runtime + doc comments

mattwparas:
- Contains everything from #47 
- Adds `call/cc`, as well as all the tools necessary to embed async functions inside the runtime, and poll them to completion
- Using `call/cc`, we can create an async runtime as well.


tov:
Whoa.

Issue #65, opened 2023-06-20
Add case insensitive string primitives


mattwparas:
All of the current string primitives are case sensitive - add the equivalent `-ci` versions as well, see [the racket docs](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3d~3f%29%29) for inspiration.

Start [here](https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/primitives/strings.rs)

In particular, these functions:

[string=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3d~3f%29%29)
[string<?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3c~3f%29%29)
[string<=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3c~3d~3f%29%29)
[string>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3f%29%29)
[string>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3d~3f%29%29)
[string-ci=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3d~3f%29%29)
[string-ci<?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3c~3f%29%29)
[string-ci<=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3c~3d~3f%29%29)
[string-ci>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3f%29%29)
[string-ci>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3d~3f%29%29)

mattwparas:
Added in #81

PR #81, opened 2023-09-20
Case insensitive string functions

rktjmp:
Re https://github.com/mattwparas/steel/issues/65

Does not include [string-ci>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3f%29%29), [string-ci>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3d~3f%29%29) because Steel does not include [string>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3f%29%29),[string>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3d~3f%29%29). I can add them if you want?

I was originally deferring the case insensitive checks to the non-ci versions, but since they accept a `SteelString` and `to_lowercase()` returns a `String`, it seemed not worth re-wrapping the types (?) and I just repeated the comparison code.

### DIFF
--- a/cogs/r5rs.scm
+++ b/cogs/r5rs.scm
@@ -373,6 +373,14 @@
 (check-equal? "string <=, true" #t (string<=? "a" "aa"))
 (check-equal? "string <=, same string" #t (string<=? "a" "a"))
 
+(check-equal? "case-insensitive string-equality with constructor, equal" #t (string-ci=? "A" (string #\a)))
+(check-equal? "case-insensitive string-equality with constructor, not equal" #f (string-ci=? "A" (string #\b)))
+(check-equal? "case-insensitive string<, true" #t (string-ci<? "A" "aa"))
+(check-equal? "case-insensitive string<, false" #f (string-ci<? "AA" "a"))
+(check-equal? "case-insensitive string<, same strings" #f (string-ci<? "A" "a"))
+(check-equal? "case-insensitive string <=, true" #t (string-ci<=? "A" "aa"))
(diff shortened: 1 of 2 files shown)


rktjmp:
Though re-reading this, probably it is best to rewrap `to_lowercase() -> SteelString` so if `<` etc is ever overloaded we pickup the behaviour?

mattwparas:
> Does not include [string-ci>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3f%29%29), [string-ci>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string-ci~3e~3d~3f%29%29) because Steel does not include [string>?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3f%29%29),[string>=?](https://docs.racket-lang.org/reference/strings.html#%28def._%28%28quote._~23~25kernel%29._string~3e~3d~3f%29%29). I can add them if you want?

If you could add those as well that would be great! Also per the second point, `SteelString` is really just a wrapper around `Rc<String>` - so I think we are safe to just use the `to_lowercase() -> String` - converting to the `SteelString` I think won't be necessary for our case

mattwparas (approved):

mattwparas:
closes #65

rktjmp:
Cheers for the quick shepherding.

PR #82, opened 2023-09-20
Fix comments referencing $STEEL_PATH instead of $STEEL_HOME

rktjmp:
Extremely minor, but seems the env var name changed at some point.


mattwparas (approved):

mattwparas:
Thanks!

Issue #84, opened 2023-09-21
testing system and enahnce command `steel test`


AngeloChecked:
I attempted to understand the inner workings of `run-tests.scm` [here](https://github.com/mattwparas/steel/blob/master/cogs/test-runner.scm).

Today, we could consider placing a default copy of `run-tests.scm` in the project directory, allowing `steel test` to run that file as one possible option.

I believe the purpose of the `steel test` command is to simplify test execution for users, abstracting away the details of the testing system implementation. This implies that we would need to reimplement `run-test.scm` within the command line. However, this approach has a tradeoff – it makes 'steel' responsible for the testing strategy. I'm not entirely sure how other programming languages handle different testing frameworks (like JUnit in Java, Mocha and Jest in JavaScript, or HSpec, HUnit, QuickCheck... in Haskell). They either have standalone command-line interfaces or provide a test driver of some kind. I think Golang `go test` and Rust `cargo test` are very opinionated here. But I admit I'm not well-informed on this topic.

as an initial proof of concept I implemented it by replicating the behavior of  `run-tests.scm` as follows:
<details>
<summary>rust example</summary>

```rust
fn test_files(dir: &PathBuf, paths: &mut Vec<PathBuf>) -> Result<(), Box<dyn Error>> {
    if dir.is_dir() {
        for entry in fs::read_dir(dir)? {
            let entry = entry?;
            let path = entry.path();
            if path.is_dir() {
                test_files(&path, paths)?;
            } else {
                if path.is_file()
                    && path
                        .file_name()
                        .map(|name| name.to_str().map_or(false, |n| n.ends_with(".scm")))
                        .unwrap_or(false)
                {
                    paths.push(path);
                }
            }
        }
    }
    Ok(())
}

fn run_folder_tests(project_root: &PathBuf) -> Result<(), Box<dyn Error>> {
    let mut vm = Engine::new();
    vm.register_value("std::env::args", steel::SteelVal::ListV(vec![].into()));
    vm.run("(set-test-mode!)")?;

    let mut files = Vec::new();
    if project_root.is_file() {
        files.push(project_root.clone());
    } else {
        test_files(project_root, &mut files)?;
    }

    for file in &files {
        let contents = fs::read_to_string(&file)?;
        if let Err(e) = vm.compile_and_run_raw_program_with_path(&contents, file.clone()) {
            e.emit_result(file.to_str().unwrap(), &contents);
            return Err(Box::new(e));
        }
    }
[...]

PR #95, opened 2023-11-05
implementation of command test

AngeloChecked:
image
image

The "test command" serves to run the test runner and gather feedback from the test suite concerning the overall status of our program.

Different programming languages and ecosystems use a range of approaches to tackle this task. Some use external tools, while others integrate it into their toolchains. Certain ecosystems rely on third-party libraries, and there are those that aim to standardize the process as much as possible.

This current implementation is limited to utilizing/embedding ([this](https://github.com/mattwparas/steel/blob/master/cogs/test-runner.scm)) the existing testing system and providing a user-friendly interface through the Steel CLI. This enables running tests in the module's working directory or choosing specific tests to run.

we can using the command in this ways:
```sh
steel test   # run the tests in the current directory
steel test . # equal to above
steel test contracts/contract-test.scm # run only test of the selected file
```

### note
- To make this work, I had to adjust the import path of  [test-runner.scm](https://github.com/mattwparas/steel/blob/master/cogs/test-runner.scm), but it doesn't appear to be a concern.
- To ensure that the external file test works correctly (now they must `provide` something to works), we must address the tree-shaking issue we previously discussed #84 .

I welcome all feedback and suggestions for improvement

### DIFF
--- a/cogs/test-runner.scm
+++ b/cogs/test-runner.scm
@@ -1,4 +1,4 @@
-(require "fs/fs.scm")
+(require "steel/fs/fs.scm")
 
 (define shared-engine (Engine::new))
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@ enum EmitAction {
         default_file: Option<PathBuf>,
         arguments: Vec<String>,
     },
-    /// Test the module
+    /// Tests the module - only tests modules which provide values
     Test { default_file: Option<String> },
     /// Generate the documentation for a file
     Doc { default_file: Option<PathBuf> },
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,12 +95,24 @@ pub fn run(clap_args: Args) -> Result<(), Box<dyn Error>> {
 
         Args {
             default_file: None,
-            action: Some(EmitAction::Test { default_file: _ }),
+            action: Some(EmitAction::Test { default_file }),
             ..
         } => {
-            todo!()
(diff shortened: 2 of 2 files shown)


mattwparas:
This looks good! Per needing to have a `provide` - I have addressed the tree shaking issue - now the test runner actually only calls `require` on files that actually have a `provide`. Reason being, I had no other way to delineate an application `.scm` file from a file meant to be used as a library, without consulting some other kind of file such as the `cog.scm`. For the first iteration here though, I think the existing behavior is fine as is as long as we document that this will only attempt to require files with provides.

Any thoughts on that?

AngeloChecked:
> This looks good! Per needing to have a `provide` - I have addressed the tree shaking issue - now the test runner actually only calls `require` on files that actually have a `provide`. Reason being, I had no other way to delineate an application `.scm` file from a file meant to be used as a library, without consulting some other kind of file such as the `cog.scm`. For the first iteration here though, I think the existing behavior is fine as is as long as we document that this will only attempt to require files with provides.
> 
> Any thoughts on that?


Yep, for the initial iteration, I believe this is good enough. 

in your opinion, should we consider printing a message after the `steel test` to provide a warning about it?

mattwparas:
I think just add a little flavor for the doc text of steel test - so instead of just "test the module" we could include something like:

"Tests the module - only tests modules which provide values" or something of the sort 

Otherwise - looks good to me, once you make that update we can run the pipeline, I'll approve and we can merge

AngeloChecked:
done

PR #112, opened 2023-12-06
fix flake and dev shell

gabydd:
Just updates the Cargo.lock so the flake builds and updates the devshell to include openssl


mattwparas:
Thanks!

Issue #114, opened 2023-12-06
Add `make-string` primitive


mattwparas:
Missing the `make-string`  - see the spec for details https://small.r7rs.org/attachment/r7rs.pdf

If possible, identify any tests under `cogs/r5rs.scm` that are skipped currently that need this primitive, and make sure they work.

See `steel-core/src/primitives/strings.rs` for other string primitives

gabydd:
was working on this a bit, string is already implemented see here: https://github.com/mattwparas/steel/blob/4f77effec20cde870f2beae66fff8fd772bbda0c/crates/steel-core/src/primitives/strings.rs#L157-L160, I was able to implement make-string but I had to do some manual arity checking because I don't think the Range arity(which is what I assume would be used for having either 1 or 2 args is usable. Totally fine @duttakapil if you want to go ahead and work off of what I have done as I won't have time this week for this(exams). This is my initial implementation (there might be a better way to do arity that I don't know of):
```rs
#[function(name = "make-string")]
pub fn make_string(k: usize, c: RestArgs<char>) -> Result<SteelVal> {
    if c.len() > 1 {
        stop!(ArityMismatch => format!("make-string expected 1 or 2 arguments, got {}", c.len() + 1)) 
    }
    let c = c.first().unwrap_or(&'\0');
    Ok((0..k).into_iter().map(|_| c).collect::<String>().into())
}
```

mattwparas:
Good catch on the string function already existing, missed it. And yeah, this looks fine - unfortunately the multi-arity macro primitives I have are fairly naive, but your implementation looks sane enough, just have to test it. I don't remember off hand the details of RestArgs

mattwparas:
This is how you'd get this to work:
```rust
#[function(name = "make-string")]
pub fn make_string(k: usize, mut c: RestArgsIter<'_, char>) -> Result<SteelVal> {
    // If the char is there, we want to take it
    let char = c.next();

    // We want the iterator to be exhaused
    if let Some(next) = c.next() {
        stop!(ArityMismatch => format!("make-string expected 1 or 2 arguments, got an additional argument {}", next?))
    }

    let c = char.unwrap_or(Ok('\0'))?;
    Ok((0..k).into_iter().map(|_| c).collect::<String>().into())
}

```

In this case the RestArgs is just an iterator, so we check if it is exhausted to assert we have no more arguments

mattwparas:
Fixed in #116

PR #116, opened 2023-12-08
Add make-string primitive, enable associated test cases

duttakapil:
I have added the `make-string` primitive, as discussed in https://github.com/mattwparas/steel/issues/114

I have enabled the skipped tests in `cogs/r5rs.scm` as well and also added a few new test cases 

Please let me know if this is good enough

### DIFF
--- a/cogs/r5rs.scm
+++ b/cogs/r5rs.scm
@@ -422,8 +422,8 @@
 (check-equal? "case-insensitive string >=, true" #t (string-ci>=? "aa" "A"))
 (check-equal? "case-insensitive string >=, same string" #t (string-ci>=? "a" "A"))
 
-(check-equal #t (string=? "a" (make-string 1 #\a)))
-(check-equal #f (string=? "a" (make-string 1 #\b)))
+(check-equal? "make-string creates single character string 'a' correctly" #t (string=? "a" (make-string 1 #\a)))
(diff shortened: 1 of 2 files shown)


mattwparas:
This looks good - do you mind rebasing with the latest changes from master? Otherwise this test will fail

cogs/r5rs.scm
```diff
@@ -417,8 +422,8 @@
 (check-equal? "case-insensitive string >=, true" #t (string-ci>=? "aa" "A"))
 (check-equal? "case-insensitive string >=, same string" #t (string-ci>=? "a" "A"))
 
-(skip-compile (check-equal #t (string=? "a" (make-string 1 #\a)))
-              (check-equal #f (string=? "a" (make-string 1 #\b))))
+(check-equal #t (string=? "a" (make-string 1 #\a)))
+(check-equal #f (string=? "a" (make-string 1 #\b)))
```
mattwparas, line 426: These will need to be changed to be `check-equal? ` followed by a description of the test, see the other test cases for an example

duttakapil: I am thinking it's better to remove these two additional test cases that I have added, since your first test case already checks for this scenario

mattwparas: if possible, could we leave them in? It makes sure the result _isn't_ `string=?` when given a different seed character

duttakapil: I have updated the PR

mattwparas (approved):

PR #135, opened 2024-01-15
Document embedding the REPL within a Rust application

wmedrano:
Demo: https://wmedrano.github.io/steel/book/start/embedded.html


docs/src/start/embedded.md
```diff
@@ -1 +1,77 @@
...
+## Embedding The Steel REPL
+
+Repl functionality is provided by the `steel-repl` crate.
+
+```toml
+[dependencies]
+steel-repl = { git="https://github.com/mattwparas/steel.git", branch = "master" }
+```
+
+### repl_base
```
wmedrano, line 66: Is it ok to stabilize this a bit more? If so, perhaps it can be renamed to `steel_repl::run_repl` or maybe refactored to:

```rust
let repl = steel_repl::Repl::new();
repl.run().unwrap();
```

wmedrano: For reference: I think this is the Guile C equivalent: https://www.gnu.org/software/guile/manual/html_node/Initialization.html#index-scm_005fshell

mattwparas: Yeah, I agree the name isn't great as is. `run_repl` as a standalone function sounds good, and then maybe a `Repl` struct like you mentioned to configure it a bit more if needed.

Do you want to make that change as part of this? If not - I can make the adjustments in a follow up PR

docs/src/builtins/steel_strings.md
```diff
@@ -1,8 +1,11 @@
 # steel/strings
+
 #### steel/strings
 
-Strings in Steel are immutable, fixed length arrays of characters. They are heap allocated,
-and are implemented under the hood as referenced counted rust `Strings`.
+Strings in Steel are immutable, fixed length arrays of characters. They are heap
+allocated, and are implemented under the hood as referenced counted Rust
+`Strings`. Rust `Strings` are stored as UTF-8 encoded bytes.
+
```
mattwparas, line 8: Thanks for these doc changes! These builtin docs get generated from the doc comments in the rust code, can you also dupe the changes over there?

See here for example: https://github.com/mattwparas/steel/blob/4cf5e15d415c0d0d17b22e40e6c5551fb117962b/crates/steel-core/src/primitives/strings.rs#L17

Same goes for the `string-length` function.

I have a branch going parallel to this that will help in the future, you'll just need to run `cargo xtask docgen` and it'll populate the builtin docs, but for now just duping the changes over should suffice

docs/src/start/embedded.md
wmedrano: Made it a part of this change.

docs/src/builtins/steel_strings.md
wmedrano: Done.

docs/src/start/embedded.md
mattwparas: thanks!

mattwparas (approved):

PR #143, opened 2024-01-18
Move dylib documentation closer to "Embedding Steel in Rust" section.

wmedrano:
Also relabels it as "Using Rust in Steel"

Demo: https://wmedrano.github.io/steel/book/start/dylib.html


wmedrano:
2024-01-18 update: Converted back to draft until I get it working myself.

mattwparas:
When you say "get it working" do you mean getting dylibs working?

wmedrano:
Yes, doing my due diligence and actually following the guide.

Edit as of 5 minutes ago, I seem to have it working.

docs/src/start/dylib.md
```diff
@@ -19,15 +35,15 @@ This should create a directory structure as follows:
...
 ```toml
 [package]
 name = "steel-sys-info"
 version.workspace = true
 edition = "2021"
 
-
 [lib]
 name = "steel_sys_info"
 crate-type = ["cdylib"]
```
wmedrano, line 49: TODO: Highlight that crate-type is required.

docs/src/start/dylib.md
```diff
@@ -40,13 +56,15 @@ abi_stable = "0.11.1"
 sys-info = "0.9.1"
```
wmedrano, line 56: I think abi_stable is actually not needed.

Also, need to double check but steel-core might require adding "dylibs" feature.

wmedrano, line 56: The example uses `steel-core` from the workspace. However, maybe it would be best if we can point to a real version?

How long until we reach a possible stable major version? Or is there no minor/major distinction due to ABI issues?

mattwparas: I agree, would be best to point to a real version.

I can cut a new release to publish to crates.io this weekend and we can use that version here if that's alright, I think that'll be 0.6.0 after I publish. I don't think I'm quite ready to cut a 1.0 just yet, but given growing usage I don't think that is as far off as I originally thought

mattwparas: Good catch, that is right, the dylib feature is necessary. 

I also think you're right that abi stable isn't necessary for this example. More advanced usages  might, but I also need to just reexport Abi stable from steel itself so downstream consumers don't need to specify Abi stable as a dependency

wmedrano: So, it was needed by a macro.

wmedrano: Ack. Makes sense.

mattwparas: Dang, if that is the case then we definitely need to re export abi_stable and have the macro reference those directly

mattwparas: Do you by chance know which macro needed it?

wmedrano: I think it was `declare_module`.

Can't check for sure as I won't have a dev environment for at least a week.

wmedrano:
Tested the instructions again and it seems to work. Anything else I can do to push this through?

mattwparas:
LGTM - thanks!

mattwparas (approved):

Issue #144, opened 2024-01-18
Steel REPL Cursor "Bug"


partisani:
Not sure what this is about, but the title says it all, see:
![a](https://github.com/mattwparas/steel/assets/88511157/a96be2eb-acbb-4a65-beeb-9453dc89e690)

mattwparas:
Interesting - could you provide some details about your environment? Any help with trying to reproduce it?

partisani:
Windows 11, tried with/without WSL, using Windows Terminal/cmd.exe. I built the interpreter in release mode following the instructions in the readme ¯\\\_(ツ)\_/¯

mattwparas:
Thanks, I'll try to reproduce today. I'll boot up a windows machine and see what happens

mattwparas:
Would you look at that, I was able to reproduce - I'll investigate

mattwparas:
Looks to be this: https://github.com/kkawakam/rustyline/issues/562 - I can get in a fix later today

mattwparas:
The relevant line is here https://github.com/mattwparas/steel/blob/master/crates/steel-repl/src/repl.rs#L98

Specifically, changing the prompt to this on windows will work:

```rust
let mut prompt = "λ > ";
```

Might need a `to_string()` there, but can't test it quite this second. Feel free to make the change locally and see if it works for you

mattwparas:
Just pushed a commit that will fix this for windows: https://github.com/mattwparas/steel/commit/904906bdbb72a22b06630a88a2f0baf17d04ea1f

Sorry for the trouble! I'll leave this open for a few days, let me know if its fixed for you

partisani:
There was an error, i am working on a fix (just a single line, also my first contribution).

partisani:
See, in line 100 you set `prompt` to a `&str`, but in line 103 you set it to a `String`, on linux there won't be any problem, because in line 163 it's also set to a `String`, but on windows `prompt` is a `&str`, so it will give a mismatched types error.

mattwparas:
ah yes, so it does need the `to_string()`

partisani:
![a](https://github.com/mattwparas/steel/assets/88511157/45a71910-680c-43e4-8af4-0067e8ca77b1)

Yes. I really felt the need to say that it works using itself.

PR #150, opened 2024-01-28
Fix fix for input on windows

partisani:
Yes, this is a fix for #144, because there was another error, but @mattwparas couldn't test it, for any reason. It was simple and i had some spare time, so made this single line edit.


mattwparas (approved):

mattwparas:
Thanks! Appreciate the fix. I really need to get a workflow set up for windows so that I can catch these issues earlier

PR #152, opened 2024-02-02
Improve error message for symbol missing from `BuiltInModuleRepr`

wmedrano:
Also slightly simplify `maybe_get_environment_variable`.


crates/steel-core/src/steel_vm/builtin.rs
```diff
@@ -324,7 +324,15 @@ impl BuiltInModuleRepr {
 
     // This _will_ panic given an incorrect value. This will be tied together by macros only allowing legal entries
     pub fn get(&self, name: String) -> SteelVal {
-        self.values.get(name.as_str()).unwrap().clone()
+        match self.values.get(name.as_str()) {
+            Some(v) => v.clone(),
+            None => {
+                panic!(
+                    "symbol {name} not found in module {module_name}",
```
wmedrano, line 331: Does it make sense to also add a "did you mean...".

If so, is there already some existing reference to this functionality in the codebase I can take a look at?

wmedrano: Related: The `clap` crate uses version `0.10` of `strsim`. `normalized_levenshtein` seems like a good way to find the best candidate

https://docs.rs/strsim/latest/strsim/fn.normalized_levenshtein.html

wmedrano: Done, used the `strsim` crate to find closest match and wrote some unit tests to demonstrate the range.

wmedrano:
> This LGTM - any objection to the fact that it panics? I think in this path the idea was fail fast early and loudly, but I could see an argument for not doing that

Maybe. I don't think I have enough experience to make an informed opinion yet. At the moment the panic only slightly annoyed me in the repl since the following caused a crash which exits the repl:

```scheme
(#%require-dylib "libmylib" (only-in typo))
```

mattwparas:
Works for me, if you become increasingly annoyed we can adjust it to not panic. Should be an easy change and we can follow up with it if necessary, doesn't block this by any means

mattwparas (approved):

PR #153, opened 2024-02-03
Support comparison operations for BigNum.

wmedrano:


mattwparas (approved):

mattwparas:
ready to merge?

wmedrano:
Seems to work fine on the repl and the unit tests. I also noticed there were some number functions in `vm.rs`. Not sure how it is used. Should this be updated as well?

https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/steel_vm/vm.rs#L5352

mattwparas:
That is currently dead code and due for some cleanup - you can safely ignore that for now

PR #154, opened 2024-02-05
Fix single argument division for integers

wmedrano:
Before: `(/ 4) -> 0`

After: `(/ 4) -> 0.25`


crates/steel-core/src/primitives/nums.rs
```diff
@@ -421,22 +418,15 @@ pub fn special_add(args: &[SteelVal]) -> Result<SteelVal> {
     let mut sum_int: isize = 0;
     let mut sum_float = 0.0;
     let mut found_float = false;
-
-    // Hoping this does not allocate
```
wmedrano, line 425: It doesn't.

crates/steel-core/src/tests/success/math.scm
```diff
@@ -0,0 +1,41 @@
+(define-syntax assert-equal!
+    (syntax-rules ()
+      ((_ expected actual)
+       (assert! (equal? expected actual)))))
+
+;; Number types
+(assert! (not (equal? 10 10.0)))
```
mattwparas, line 7: You'll just need to add this file name here: https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/tests/mod.rs#L89 - the test files inside the crate don't get automatically picked up unlike in some other spots (they probably should though. but that is a different issue)

wmedrano: Done.

mattwparas (approved):

PR #157, opened 2024-02-07
Support scientific notation.

wmedrano:
Fixes the r5rs.scm `"string->number with scientific notation"` test: https://github.com/mattwparas/steel/blob/master/cogs/r5rs.scm#L251


mattwparas (approved):

mattwparas:
Thanks, this looks great!

PR #160, opened 2024-02-07
[DOC] Adding a brief documentation for standalone sub-section's Steel book

kenaryn:
This is a first iteration of the sub-section Steel book regarding to the standalone procedure in order to run Steel files from a terminal, either using cargo tool or independently.

This is my very first pull request in life so if something wrong or misplaced, please do alert me and I will try my best to amend it! :-)

### DIFF
--- a/docs/src/start/standalone.md
+++ b/docs/src/start/standalone.md
@@ -1,1 +1,70 @@
 # Using Steel on its own
+
+## Preambule
+This document includes some features of ABNF (Augmented Backus Naur form), which is a meta-language used to define
+another language and proves itself always useful in documentations associated with programming languages and softwares.
+To know more about its full specifications, see: https://www.rfc-editor.org/rfc/rfc5234 and its superseeded RFC at
+https://www.rfc-editor.org/rfc/rfc7405 (related to case-sensitivity).
+
+The main considerations are defined as follow:
+  - <file> = defines a logical entity between angle brackets, with a concept's name or identifier
+  - <cmd-1> / <cmd-2> = a *solidus* (as named by the Unicode consortium, more commonly known as "slash") stands for "alternative
+  OR (also known as XOR, or exlusive OR) and is inserted between to operands, forms, or S-expressions.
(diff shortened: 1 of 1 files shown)


docs/src/start/standalone.md
```diff
@@ -1 +1,85 @@
...
+  ```
+  set paths = [
+    ~/\<your/path/to/steel/directory\>/target/release`
+  $@paths]
+  ```
+  
+  5. Then run `exec elvish` and your are good to go! You can check your newly added path by typing:\
+  `pprint $paths`
+
+    - For those who still live in the dark caves of ancient shells like Bash or Zsh, simply feed the new path like the
```
wmedrano, line 71: IMO: I would move bash and zsh up as they have higher adoption. Also remove any language that makes any judgement on which is better to avoid alienating any users.

docs/src/start/standalone.md
```diff
@@ -1 +1,85 @@
...
+  `fish_add_path \<your/path/to/steel/directory\>/target/release`
+
+    - `sh`, `bash`, `zsh` and `fish` users now have to regenerate their environment variables by running:\
+  `source ~/.bashrc` / `source ~/.config/fish/config.fish` (depending on your kryptonite)
+
+
+## Running a Steel file
+
+  6. Finally, compile and execute your Steel files from your terminal like so:\
+  `steel \<steel_file.scm\>`
```
wmedrano, line 85: Can you add another section for how to install the built in modules?

The instructions right now are encoded in the CI: https://github.com/mattwparas/steel/blob/master/.github/workflows/rust.yml#L44

kenaryn: I tried what I understood but some concepts are beyond me. What is a cog? Why adopting `.steel` as a default installation?
Do I had to add all `install` related commands?

Also, I clicked the `Sync fork` button in my personal repository then chose the `Update branch` option. Will it cause some inconveniences to target the upstream repository?
I am a very beginner so I am straightforward even if as I could pass as a silly person doing so.

PR #161, opened 2024-02-08
Add more number operations.

wmedrano:
Taken from section 6.2 of https://small.r7rs.org/attachment/r7rs.pdf

New additions are: `number?`, `real?`, `rational?`
Fixed: `int?` and `integer?`

### DIFF
--- a/crates/steel-core/src/steel_vm/primitives.rs
+++ b/crates/steel-core/src/steel_vm/primitives.rs
@@ -5,7 +5,6 @@ use super::{
     register_fn::RegisterFn,
     vm::{get_test_mode, list_modules, set_test_mode, VmCore},
 };
-use crate::values::lists::List;
 use crate::{
     gc::Gc,
     parser::{
@@ -18,7 +17,10 @@ use crate::{
         hashmaps::{HM_CONSTRUCT, HM_GET, HM_INSERT},
         hashsets::hashset_module,
         lists::{list_module, UnRecoverableResult},
-        nums::{quotient, ADD_PRIMITIVE_DEFINITION},
+        nums::{
+            quotient, ADD_PRIMITIVE_DEFINITION, DIVIDE_PRIMITIVE_DEFINITION, INEXACTP_DEFINITION,
+            MULTIPLY_PRIMITIVE_DEFINITION, SUBTRACT_PRIMITIVE_DEFINITION,
+        },
         port_module,
         process::process_module,
         random::random_module,
@@ -48,6 +50,7 @@ use crate::{
         },
     },
 };
+use crate::{primitives::nums::EXACTP_DEFINITION, values::lists::List};
 use crate::{
     rvals::IntoSteelVal,
     values::structs::{build_option_structs, build_result_structs},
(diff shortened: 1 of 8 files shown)


crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -671,24 +674,51 @@ fn not(value: &SteelVal) -> bool {
...
 
 #[steel_derive::function(name = "float?", constant = true)]
 fn floatp(value: &SteelVal) -> bool {
     matches!(value, SteelVal::NumV(_))
 }
 
-#[steel_derive::function(name = "number?", constant = true)]
-fn numberp(value: &SteelVal) -> bool {
-    matches!(value, SteelVal::NumV(_) | SteelVal::IntV(_))
+#[steel_derive::function(name = "real?", constant = true)]
```
wmedrano, line 704: What's the preferred way of registering these functions? I tried to standardize on the `steel_derive` macros since I kind of like them.

crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -790,11 +820,12 @@ fn identity_module() -> BuiltInModule {
     module
         // .register_value("int?", gen_pred!(IntV))
         .register_native_fn_definition(NOT_DEFINITION)
-        .register_native_fn_definition(INTEGERP_DEFINITION)
+        .register_native_fn_definition(NUMBERP_DEFINITION)
         .register_native_fn_definition(INTP_DEFINITION)
+        .register_native_fn_definition(INTEGERP_DEFINITION)
         .register_native_fn_definition(FLOATP_DEFINITION)
-        .register_native_fn_definition(NUMBERP_DEFINITION)
```
wmedrano, line 796: Is there a way to (unit) test that a symbol isn't registered more than once?

crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -671,24 +674,51 @@ fn not(value: &SteelVal) -> bool {
...
 }
 
-#[steel_derive::function(name = "number?", constant = true)]
-fn numberp(value: &SteelVal) -> bool {
-    matches!(value, SteelVal::NumV(_) | SteelVal::IntV(_))
+#[steel_derive::function(name = "real?", constant = true)]
+fn realp(value: &SteelVal) -> bool {
+    matches!(
+        value,
+        SteelVal::IntV(_)
```
wmedrano, line 708: I'm thinking of some renames to match the r7rs language more. Should be really easy with `rust-analyzer`. LMK which ones you think are worth doing.

```
IntV -> Int
BigNum -> BigInt
FractV -> Rational or maybe Rat
BigFractV -> BigRational or maybe BigRat
NumV -> Real
```

crates/steel-core/src/steel_vm/primitives.rs
mattwparas: The derive macros are good! They're newer so some of the older functions don't have them. Doc comments on them automatically make it in to the docs so they're preferred when possible

crates/steel-core/src/steel_vm/primitives.rs
mattwparas: I'm okay with the changes, Rational and BigRational are fine. My only concern is that with the enum being public the rename is a breaking change. I'm fine with the breaking change but we'd just have to bump the version numbers since its still pre 1.0

crates/steel-core/src/steel_vm/primitives.rs
mattwparas: Could do something within the `register_native_fn_definition` or `register_value` etc functions that is something like:

```rust
#[cfg(debug_assertions)]
assert!(!module.contains(symbol));
```

Just hand waving a bit, but basically in debug mode assert that the symbol isn't already registered in the module

wmedrano: Ack.

crates/steel-core/src/steel_vm/primitives.rs
wmedrano: Ack. Updated just `Fract` since that's the one that's annoying me the most.

wmedrano:
I think this is good to go.

crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -671,24 +674,65 @@ fn not(value: &SteelVal) -> bool {
...
+    )
+}
+
+/// Returns #t if obj is a rational number, #f otherwise.
+/// Rational numbers are numbers that can be expressed as the quotient of two numbers.
+/// For example, 3/4, -5/2, 0.25, and 0 are rational numbers, while
+/// Example:
+/// ```lisp
+///   (rational? (/ 0.0)) ⇒ #f
+///   (rational? 3.5)     ⇒ #t
```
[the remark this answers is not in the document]
mattwparas: I think it makes sense, if it can be coerced to a rational type, but yes I agree that is surprising

mattwparas (approved):

PR #162, opened 2024-02-09
Performance improvements

mattwparas:
A whole host of changes:
* Bug fixes
* Adding some new opcodes which improve VM performance
* Upgrade list library version
* Change lists to use VLists, which exponentially grow in size rather than keeping a fixed size. The size starts at 4 and grows by a factor of 2 each time a new node is added.


Cargo.toml
```diff
@@ -42,6 +42,6 @@ members = [
 ]
 
 [profile.release]
-debug = false
-# debug = true
+# debug = false
```
mattwparas, line 45: will get removed

crates/steel-core/src/core/instructions.rs
```diff
@@ -117,11 +117,13 @@ pub fn disassemble(instructions: &[Instruction]) -> String {
 pub struct DenseInstruction {
     pub op_code: OpCode,
     // Function IDs need to be interned _again_ before patched into the code?
-    pub payload_size: u32,
+    // Also: We should be able to get away with a u16 here. Just grab places where u16
+    // won't fit and convert to something else.
+    pub payload_size: u16,
```
mattwparas, line 122: This needs to be changed with a bunch of support during code gen, and in the VM loop as well

crates/steel-core/src/env.rs
```diff
@@ -38,15 +38,48 @@ impl Env {
...
-    ///
-    /// if found, return that value
-    ///
-    /// Otherwise, error with `FreeIdentifier`
     #[inline(always)]
     pub fn repl_lookup_idx(&self, idx: usize) -> SteelVal {
-        self.bindings_vec[idx].clone()
+        {
+            let ref this = self.bindings_vec[idx];
+            match this {
```
mattwparas, line 45: For whatever reason, there was performance improvements by inlining this

wmedrano: How does manual inline compare to defining `Clone` with `#[inline]`?

Is it worth looking into [profile guided optimizations](https://doc.rust-lang.org/rustc/profile-guided-optimization.html) or do you think it will add complexity to optimizing?

mattwparas: good question, profile guided optimizations is definitely something I should try

I'll check the inline clone as well

mattwparas: well that made a huge difference, using `cargo pgo` I shaved like 30% off of the nqueens benchmark

mattwparas: Without pgo:
```
λ (nqueen.scm) > (nqueens 13)
=> 73712
Time taken: 32.613855536s
```

With pgo:
```
λ > (nqueens 13)
=> 73712
Time taken: 23.868467826s
```

mattwparas: So I think I'll need to create a set of representative input programs for PGO to reliably use it, but overall that is very promising.

mattwparas: Confirmed that the clone really don't make much of a difference when using PGO. I don't love it anyway, so for now I just took it out and will look into how to add PGO more robustly when building for distribution.

PR #163, opened 2024-02-09
Add datastructure support for complex numbers.

wmedrano:
Follow ups: Add lexer support.


wmedrano:
Still need to fix the equality and ord operations.

wmedrano:
If tests pass, I think this is good to go. I'll follow up with the lexer.

Also, I think I'll write some tests according to my reading of section `6.2. Numbers` of the r7rs doc. The closest I found is Chibi's tests which I think are already tested by Steel.

mattwparas (approved):

PR #166, opened 2024-02-11
Support +inf.0 -inf.0 +nan.0 -nan.0

wmedrano:
Required storing a "next" token since `+inf.0` was being parsed as 2 tokens: `+` and `inf.0`.

Seems like this behavior is used to make creating addition functions easy like:

```lisp
(define x 1234)
+x => function that adds 1234.
```


crates/steel-core/src/rvals/cycles.rs
```diff
@@ -216,54 +174,13 @@ impl CycleDetector {
...
-            BoxedIterator(_) => write!(f, "#<iterator>"),
-            Boxed(b) => write!(f, "'#&{}", b.borrow()),
-            Reference(x) => write!(f, "{}", x.format()?),
             HeapAllocated(b) => write!(f, "'#&{}", b.get()),
+            _ => {
+                self.depth -= 1;
+                let res = self.format_with_cycles(val, f);
+                self.depth += 1;
+                res
+            }
```
mattwparas, line 183: I'm having some trouble following this change - what was the purpose here?

wmedrano: Redid this part, I think this is cleaner. I can also revert back to having to separate but similar functions.

- High level: format_with_cycles and top_level_format_with cycles seem to be very similar ([difference](https://www.diffchecker.com/3X0rhqAZ/)) so tried to unify.
- Previously reviewed approach: Reused one of the functions when the special cases didn't match. But this means the depth was artificially increased a level. Probably doesn't matter but tried to keep the depth anyways.
- New approach - Pass an enum specifying which behavior to use.

mattwparas: got it, I understand, I haven't looked at some of these functions in a long time :smile:

mattwparas (approved):

Issue #62, opened 2023-06-17
Proper numeric tower


mattwparas:
It would be nice to have a proper numeric tower, right now there are integers that promote to floats at overflow - but otherwise bigints are not implemented - historically this was to avoid having to box all integers.

It might be worth implementing now, and just wrap up this crate https://docs.rs/num/latest/num/

wmedrano:
What's the status of this? It seems like `BigInts` are implemented. The following seems just fine in the terminal:

```scheme
λ > (define x 10)
λ > (+ x 1000000000000000000000000000000000000000000000000000000000000000000000)
=> 1000000000000000000000000000000000000000000000000000000000000000000010
```

mattwparas:
yeah, big ints are supported now (of course, probably missing a few things here and there) but rationals are not yet supported. That would be next on the list

wmedrano:
For implementation, what do you think of adding another variant to `SteelVal`:

```rust
enum Fraction {
    Small(Rational32),
    Big(BigRational),
}

enum SteelVal {
    ...
    FractV(Fraction),
    ...
}

// We should stay under this constraint when adding a new type.
const _ASSERT_SMALL: () = assert!(std::mem::size_of::<SteelVal>() <= 16);
```

If it looks alright, I may try prototyping it.

mattwparas:
This looks alright - unfortunately you might have to `Gc` it like how `BigNum` is:

```rust
BigNum(Gc<num::BigInt>),
```

the rational 32 might be able to be unboxed, however to do that you'd have to do lift the enum variant up I think unless some enum variant size optimization kicks in.

If this works, that is pleasant:

```rust
enum Fraction {
    Small(Rational32),
    Big(Gc<BigRational>),
}

enum SteelVal {
    ...
    FractV(Fraction),
    ...
}
```

However you might have to do this:

```rust
enum SteelVal {
    ...
    SmallFract(Rational32),
    BigFract(Gc<BigRational>),
    ...
}
```

mattwparas:
Otherwise - definitely give prototyping it a shot

wmedrano:
Is the goal to implement the full numerical tower? If so, do you have a checklist of requirements? From skimming [Wikipedia](https://en.wikipedia.org/wiki/Numerical_tower) it seems like `complex` numbers are the only thing left.

mattwparas:
This could be a general reference: https://www.gnu.org/software/guile/manual/html_node/Numerical-Tower.html

[...]

PR #174, opened 2024-03-02
Support exact-integer-sqrt

wmedrano:
Part of #62


wmedrano:
I think this is good if the tests pass.

mattwparas (approved):

mattwparas:
Looks good, thanks!

PR #178, opened 2024-03-06
Improve development experience on nix

sirius94:


mattwparas:
Sorry for the late reply - thank you!

PR #180, opened 2024-03-07
core: Fix panic on division by integer 0

sirius94:
Division by `0isize` now behaves the same as division by `0f64`. Both produce `+inf.0`.

```scheme
(= (/ 1 0) (/ 1 0.0)) ; => #true
```


mattwparas (approved):

mattwparas:
Thanks!

PR #181, opened 2024-03-07
core: Support path expansion (#87)

### LINKED ISSUE #87: Support path expansion
I would like to be able to specify paths starting from `~` for example: `~/.local/` and it'll expand to `/home/user/.local`
This could be useful for interacting with other CLI tools or saving some state/data to a specific folder in the user directory.

Maybe the usage could be something like this(sorry my Scheme is nonexistant)

```scheme
(expand-path "~/.local/state/helix") ;; evals to "/home/user/.local/state/helix" 
```

sirius94:
`std::fs::canonicalize` does most of the work, but it doesn't expand `~`. `std::env::home_dir` is deprecated. So I added the `dirs` dependency, which is relatively light and only has one dependency.

Features:
- `.` and `..` are resolved
- `~` is resolved to the current users home directory (should work on all major OSs; see the documentation for `dirs::home_dir`)
- symbolic links are resolved (see `std::fs::canonicalize`)

Limitations:
- expansion of other users home directories is not supported (~user/)
- expansion only works for files which exist on the filesystem (see `std::fs::canonicalize`)

Closes #87

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -945,7 +945,16 @@ version = "4.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ca3aa72a6f96ea37bbc5aa912f6788242832f75369bdfdadcb0e38423f100059"
 dependencies = [
- "dirs-sys",
+ "dirs-sys 0.3.7",
+]
+
+[[package]]
+name = "dirs"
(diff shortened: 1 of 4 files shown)


mattwparas (approved):

mattwparas:
Thanks!

PR #195, opened 2024-04-26
Add Lisp introduction for the book, part 1

fominok:
Hello @mattwparas,

First of all, I'd like to thank you for your efforts in making embedded Lisp for Rust, as well as advocating for it through the actual implementation of a plugin system for Helix.

I've been observing responses from individuals who are resistant to the idea of Lisp as an embedded language for the editor, and in response I've taken it upon myself to write an introductory Lisp chapter for the book. My hope is that this will help broaden the reach of the project and make it more approachable.

While my work is not yet finished, I wanted to share my progress early. If this idea resonates with you, I am eager to continue my efforts.


mattwparas:
Thanks for the PR - I will take a look in the next few days

docs/src/lisp/bindings_scopes.md
```diff
@@ -0,0 +1,81 @@
...
+`let` is used as follows: `(let ((symbol-a value-a) (symbol-b value-b) ...) body)`.
+
+```lisp
+λ > (let ((a (* 25 4))
+          (b 3))
+      (* a b))
+=> 300
+λ > a
+=> 1337
+```
```
mattwparas, line 71: In steel, `define` desugars to `let` when used in a local scope, so I think this distinction should be made here. In fact, I'd argue that define is preferred to let stylistically

docs/src/lisp/overview.md
```diff
@@ -0,0 +1,20 @@
...
+indentation. However, in Lisp, we gain a beautiful property: while editing and navigating code,
+a programmer can operate on a syntax tree rather than individual characters and words. In fact,
+in Lisp, every expression can be traversed with proper editor support &mdash; moving forward, backward,
+to the parent, and to nested expressions, opposed to characters, words or in the best case type
+definitions and functions in other languages. This level of structural clarity and manipulation is a
+testament to Lisp's elegant design and contributes to its expressiveness and flexibility.
+
+And since we're in software engineering, let your curious mind enjoy something new and hopefully
+rewarding, or at least look at how many smileys you can put at the end of a function with all these
+closing parentheses!
```
mattwparas, line 20: While I appreciate the sentiment of this, I fear that this comes across as a little preachy. In my personal opinion, I don't think we need to put down other languages in any capacity

docs/src/lisp/bindings_scopes.md
fominok: Could you please provide an example? I'm not so familiar with steel and learned it as I wrote this, I have only elisp and clojure experience to rely on

fominok: I tried to use `define` in a local scope as far as I understood it correctly: 
```
λ > (let ((a 5)) (define b 4) (+ a b))
=> 9
λ > b
=> 4
```

And it escaped; was it a local scope you mentioned and this is a bug?

mattwparas: That is a bug! Good catch, look like this is a regression and I missed a test. I'll get a fix up

What I was referring to is like this:

```scheme
(define (test) (define foo 10) (+ foo 100))
```

mattwparas: ah, its a bug with const evaluation. See that this works fine:

```scheme
(let ((foo (list 10 20 30 40))) (define bar (list 50 60)) (append foo bar))
```

mattwparas: Will get patched by this #200

fominok: I've updated the chapter with both ways of defining symbols locally, please take a look

docs/src/lisp/overview.md
fominok: I agree, tried to make it less preachy now, yet showing some benefits to use it

docs/book.toml
```diff
@@ -1,13 +1,6 @@
...
 multilingual = false
 src = "src"
-title = "Steel"
-
-# https://github.com/slowsage/mdbook-pagetoc
-[preprocessor.pagetoc]
-
-[output.html]
-additional-css = ["theme/pagetoc.css"]
-additional-js  = ["theme/pagetoc.js"]
```
mattwparas, line 13: Could we keep this in?

mattwparas (changes requested):
Would it be possible to make the distinction between Scheme and Lisp? I don't mind referring to steel as a member of the lisp family (since it is) but in general its origin bears much closer resemblance to scheme, and I'd like that to be reflected in the documentation

docs/book.toml
fominok: Ah, I didn't figure out what was that artifact and removed it

it was just an empty TOC, I addressed that in .css

fominok:
> Would it be possible to make the distinction between Scheme and Lisp? I don't mind referring to steel as a member of the lisp family (since it is) but in general its origin bears much closer resemblance to scheme, and I'd like that to be reflected in the documentation

hey, I tried to frame it more as a descendant of Scheme

docs/src/foundations/bindings_scopes.md
```diff
@@ -0,0 +1,118 @@
...
+result in an error and make no sense), the _value_ part follows standard rules.
+
+### `let` and local scope `define`
+
+If `define` seems reminiscent of global variables to you, you're not alone. While it's suitable for
+defining function bindings (more on that later) or constants, there must be a better way for local
+bindings.
+
+Steel has two ways to do it: one is a Lisp classic, `let` bindings, and another one is
+Steel-specific local scope `define`.
```
mattwparas, line 62: Not a deal breaker, but this isn't specific to steel, local defines are used in both Racket and Guile - as well as other schemes. I'm not 100% sure if it is part of the spec itself, I'd have to check.

docs/src/introduction.md
```diff
@@ -0,0 +1,20 @@
+# Introduction
+
+Welcome to "The Book of Steel" &mdash; a guide for a dynamic and versatile programming language
```
mattwparas, line 3: A nit, could we go with `The Steel Book` instead of the `The Book of Steel`? Happy to discuss this

docs/src/foundations/bindings_scopes.md
fominok: You're right, can see such examples in r6rs

docs/src/introduction.md
fominok: Renamed, it was the first thing that came to my mind to name it, eventually it's up to you

docs/src/introduction.md
```diff
@@ -0,0 +1,20 @@
...
+
+Rust provides robustness and safety, ensuring reliable and efficient software development. However,
+its rigidity can sometimes feel limiting when facing dynamic and evolving requirements, such as
+those encountered in a text editor, a crucial tool that developers tailor to their needs. Enter
+Lisp &mdash; a language family renowned for its extensibility and dynamic nature. It serves as the
+ultimate extension language, designed to be extendable itself by allowing developers to create DSLs
+and tweak the evaluation process to fit any domain or task, even those unforeseen or irrelevant to
+the original application.
+
+In "The Book of Steel," our aim is twofold: to demystify language basics for developers unfamiliar
```
mattwparas, line 18: Nit - this should now be `"The Steel Book"`

mattwparas:
Just the one nit, otherwise I think this looks okay - did you want to get this merged as is? Or continue adding things?

docs/src/introduction.md
fominok: oops, thanks

fominok:
> Just the one nit, otherwise I think this looks okay - did you want to get this merged as is? Or continue adding things?

Thanks for the review, I planned to do more PRs to make them more manageable as I'm not quite sure where is the point when I find it "done", so this one could be merged

mattwparas (approved):

Issue #192, opened 2024-04-23
Dependency `lasso` fails to compile on nightly toolchain


tychedelia:
See https://github.com/Kixiron/lasso/pull/45. Maintainer appears to be MIA. Looked briefly at Steel's use and it seems non-trivial to factor out.

tychedelia:
Here's the patch from https://github.com/Kixiron/lasso/pull/45/files for anyone who needs it

```toml
[patch.crates-io]
lasso = { git = "https://github.com/lcnr/lasso/", branch = "patch-1" }
```

mattwparas:
Oh that's unfortunate, I hope Kixiron is alright - he's put out some very impressive work.

I'll take a look at this more closely later today.

mattwparas:
I've patched master now to have the patched version of lasso

amkhlv:
please update [steel-core](https://crates.io/crates/steel-core) to use the latest [v7.2 of lasso](https://github.com/Kixiron/lasso)

mattwparas:
It is updated on master, are you asking for a new release?

amkhlv:
I am sorry about this confusion. But the current release just does not build...

mattwparas:
What are you trying to build? Can you provide any more details?

mattwparas:
The current release of steel on crates will have issues on newer rust versions, due to the issue with lasso. If you take a dependency on the git version of steel on the latest master, that should build

I will address the crates release of steel soon

amkhlv:
I tried git version. I my Cargo.toml :

```
steel-core = { git = "https://github.com/mattwparas/steel" }
```

then I am getting:

```
error: failed to select a version for `getrandom`.
    ... required by package `steel-core v0.6.0 (https://github.com/mattwparas/steel#1eb4d9b4)`
```
```
the package `steel-core` depends on `getrandom`, with features: `js` but `getrandom` does not have these features
```

mattwparas:
are you compiling for wasm?

amkhlv:
No. It is a big sloppy personal utilities collection:

[utils/Cargo.toml](https://github.com/amkhlv/usr/blob/master/share/Rust/utils/Cargo.toml)

mattwparas:
Interesting, I was able to reproduce this with your project, but not on a plain project. Taking a look now

mattwparas:
[...]

PR #197, opened 2024-05-02
Fetch version from cargo for repl startup msg

glenn-m:
Sources package version from Cargo instead of a static string in the REPL startup banner.

### DIFF
--- a/crates/steel-repl/src/repl.rs
+++ b/crates/steel-repl/src/repl.rs
@@ -30,6 +30,24 @@ fn display_help() {
     );
 }
 
+fn display_startup() {
+    println!(
+        "{}",
+        format!(
+            r#"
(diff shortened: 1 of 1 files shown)


glenn-m:
I believe this is failing due to https://github.com/mattwparas/steel/issues/192

mattwparas:
> I believe this is failing due to #192

ah that is a bummer isn't it. I'll take a look later today but these changes lgtm. I'll attempt to fix the build as well

mattwparas:
I've patched master, if you rebase, the build should pass

glenn-m:
> I've patched master, if you rebase, the build should pass

Perfect, thank you! I've rebased.

mattwparas (approved):

PR #199, opened 2024-05-04
Load and save history on repl start/exit

glenn-m:
Adds functionality to load/save REPL history, it defaults to the path `~/.steel/history` unless `STEEL_HOME` is set.

### DIFF
--- a/crates/steel-repl/src/repl.rs
+++ b/crates/steel-repl/src/repl.rs
@@ -19,6 +19,8 @@ use std::time::Instant;
 
 use std::env;
 
+use std::fs::File;
+
 use dirs;
 
 use crate::highlight::RustylineHelper;
@@ -138,9 +140,11 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
 
     // Load repl history
     let history_path = get_repl_history_path();
-    if let Err(err) = rl.load_history(&history_path) {
-        eprintln!("Failed to load REPL history: {}", err);
-    }
+    if let Err(_) = rl.load_history(&history_path) {
+        if let Err(_) = File::create(&history_path) {
+            eprintln!("Unable to create repl history file {:?}", history_path)
+        }
+    };
 
     let current_dir = std::env::current_dir()?;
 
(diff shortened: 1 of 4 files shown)


crates/steel-repl/src/repl.rs
```diff
@@ -113,6 +136,12 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
     let mut rl = Editor::<RustylineHelper>::new().expect("Unable to instantiate the repl!");
     rl.set_check_cursor_position(true);
 
+    // Load repl history
+    let history_path = get_repl_history_path();
+    if let Err(err) = rl.load_history(&history_path) {
+        eprintln!("Failed to load REPL history: {}", err);
+    }
```
mattwparas, line 143: If this file doesn't exist yet, what happens? Will it say that it fails to load? 

Subsequently, will it create the file when attempting to save it the history later on?

glenn-m: Right now it just errors that it cannot open it and continues. I can attempt to write the file if it doesn't exist?

```
     _____ __            __
    / ___// /____  ___  / /          Version 0.6.0 
    \__ \/ __/ _ \/ _ \/ /           https://github.com/mattwparas/steel
   ___/ / /_/  __/  __/ /            :? for help
  /____/\__/\___/\___/_/
    
Failed to load REPL history: No such file or directory (os error 2)
λ > (+ 2 2)
=> 4
λ > 
```

mattwparas: I think attempting to write the file is probably fine. At some point it might be good to implement some kind of max history length or something like that, not sure if the rustyline library supports that out of the box

glenn-m: @mattwparas It will now create the history file if it does not exist. It looks like the rustyline library does support defining a max history size, the default is 100 lines. We could make it user definable, I'm just not sure where to put that. An env var?

mattwparas: Env var is fine for now until a more thorough repl configuration is in place since there hasn't been a need for other configuration up to this point. Maybe just like `STEEL_REPL_HISTORY_LENGTH` or something? And just default to 100 otherwise

Alternatively, we could just leave it at 100 and not make it configurable for now

mattwparas (approved):

PR #203, opened 2024-05-07
Various Nix flake fixes

glenn-m:
* Fix to make `cargo test --all` work in nix shell.
* Fix to allow tests to run successfully with Nix on MacOS
* Add `rustfmt` to Nix shell so that `cargo fmt` works.
* Temporarily patch `lasso` hash until fixes are upstreamed.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,13 @@
           pname = manifest.package.name;
           version = manifest.workspace.package.version;
           src = gitignoreSource ./.;
-          cargoLock.lockFile = ./Cargo.lock;
+          cargoLock = {
+            lockFile = ./Cargo.lock;
+            # Temporary fix until https://github.com/mattwparas/steel/issues/192 is fixed upstream
+            outputHashes = {
+              "lasso-0.7.2" = "sha256-ibpHfge3nEtwLNghKEQT7ZpTe5kgDf8hbBb9qYHyHcQ=";
+            };
+          };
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -26,7 +26,13 @@
           pname = manifest.package.name;
           version = manifest.workspace.package.version;
           src = gitignoreSource ./.;
-          cargoLock.lockFile = ./Cargo.lock;
+          cargoLock = {
+            lockFile = ./Cargo.lock;
+            # Temporary fix until https://github.com/mattwparas/steel/issues/192 is fixed upstream
+            outputHashes = {
+              "lasso-0.7.2" = "sha256-ibpHfge3nEtwLNghKEQT7ZpTe5kgDf8hbBb9qYHyHcQ=";
```
mattwparas, line 33: why is this necessary?

glenn-m: I'm not sure exactly why but the `buildRustPackage` function when using a cargo file with a dependency that points to Git it needs this output hash defined. 

Without this hash if I run `nix run` I get:

```nix
       error: No hash was found while vendoring the git dependency lasso-0.7.2. You can add
       a hash through the `outputHashes` argument of `importCargoLock`:

       outputHashes = {
         "lasso-0.7.2" = "<hash>";
       };

       If you use `buildRustPackage`, you can add this attribute to the `cargoLock`
       attribute set.
```

mattwparas (approved):

mattwparas:
thanks!

PR #205, opened 2024-05-09
fix string functions based on character offsets

jrvidal:
I noticed a couple of potential issues with string functions and multi-byte characters. I'm assuming that the affected functions are supposed to be character-based, and not byte based (unlike `string-length`).

### DIFF
--- a/cogs/r5rs.scm
+++ b/cogs/r5rs.scm
@@ -391,6 +391,8 @@
 (check-equal? "string indexing into last character" #\c (string-ref "abc" 2))
 
 (check-equal? "empty substring" "" (substring "abc" 0 0))
+(check-equal? "empty substring at the end" "" (substring "abc" 3 3))
+(check-equal? "empty substring in the middle" "" (substring "abc" 1 1))
 (check-equal? "substring just the first character" "a" (substring "abc" 0 1))
 (check-equal? "substring a larger chunk" "bc" (substring "abc" 1 3))
 (check-equal? "substring with multibyte characters" "λμ" (substring "λλμν" 1 3))
--- a/cogs/r5rs.scm
+++ b/cogs/r5rs.scm
@@ -387,15 +387,16 @@
 
 (check-equal? "string indexing into first character" #\a (string-ref "abc" 0))
 (check-equal? "string indexing with multibyte characters" #\a (string-ref "λa" 1))
-
 (check-equal? "string indexing into last character" #\c (string-ref "abc" 2))
 
 (check-equal? "empty substring" "" (substring "abc" 0 0))
 (check-equal? "empty substring at the end" "" (substring "abc" 3 3))
+(check-equal? "substring without end" "bc" (substring "abc" 1))
 (check-equal? "empty substring in the middle" "" (substring "abc" 1 1))
 (check-equal? "substring just the first character" "a" (substring "abc" 0 1))
 (check-equal? "substring a larger chunk" "bc" (substring "abc" 1 3))
 (check-equal? "substring with multibyte characters" "λμ" (substring "λλμν" 1 3))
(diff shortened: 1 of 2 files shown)


cogs/r5rs.scm
```diff
@@ -386,12 +386,14 @@
 (check-equal? "string length correctly reported for standard string" 3 (string-length "abc"))
 
 (check-equal? "string indexing into first character" #\a (string-ref "abc" 0))
+(check-equal? "string indexing with multibyte characters" #\a (string-ref "λa" 1))
```
jrvidal, line 389: I could add way more tests, but I'm not sure if this is the right place.

crates/steel-core/src/primitives/strings.rs
```diff
@@ -221,24 +221,46 @@ pub fn string_ci_equals(left: &SteelString, right: &SteelString) -> bool {
...
+        value.chars().nth(index)
+    } else {
+        None
+    };
 
-    Ok(SteelVal::CharV(value.as_str().chars().nth(index).unwrap()))
+    if let Some(ch) = res {
+        Ok(SteelVal::CharV(ch))
+    } else {
+        stop!(Generic => "string-ref: index out of bounds: index: {}, string length: {}", index, value.len());
```
jrvidal, line 233: I'm leaving `value.len()` as the "string length", even though that might come off weird in some scenarios:

```
λ > (string-ref "千葉市" 4)
error[E11]: Generic
  ┌─ :1:2
  │
1 │ (string-ref "千葉市" 4)
  │  ^^^^^^^^^^ string-ref: index out of bounds: index: 4, string length: 9
```

crates/steel-core/src/primitives/strings.rs
```diff
@@ -221,24 +221,46 @@ pub fn string_ci_equals(left: &SteelString, right: &SteelString) -> bool {
...
+    }
 }
 
 #[function(name = "substring", constant = true)]
 pub fn substring(value: &SteelString, i: usize, j: usize) -> Result<SteelVal> {
```
jrvidal, line 238: :question: Should `j` be optional? (I think it is in other Schemes)

crates/steel-core/src/primitives/strings.rs
```diff
@@ -221,24 +221,46 @@ pub fn string_ci_equals(left: &SteelString, right: &SteelString) -> bool {
...
+        stop!(Generic => "substring: index out of bounds: left bound: {}", i);
+    };
+
+    let end = if j == i {
+        start
+    } else {
+        chars.nth(j - i - 1).unwrap_or(value.len())
+    };
+
+    Ok(SteelVal::StringV(value[start..end].into()))
```
jrvidal, line 263: This is not quite correct yet, give me a sec.

crates/steel-core/src/primitives/strings.rs
mattwparas: This seems fine to me

crates/steel-core/src/primitives/strings.rs
mattwparas: Yeah it looks like it is optional in Racket - did you want to make that change here?

cogs/r5rs.scm
mattwparas: you can add tests here, or if you'd like more direct ones, you can make a file in either the `success` or `failure` directories [here](https://github.com/mattwparas/steel/tree/master/crates/steel-core/src/tests)

And then add the file name in the requisite handler [here](https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/tests/mod.rs#L52)

jrvidal: I added just a couple more.

mattwparas (approved):

PR #206, opened 2024-05-09
docs for several builtins

jrvidal:
I added a few docs for several builtins. Each commit in this PR is pretty isolated, to help with review. I'm not totally sure what's the end format we're striving for, e.g.

```
(foo n)

* n : int?
```

vs.

```
(foo int?)
```

so I've winged it a little bit.

:question: Should I run `docgen` and commit the generated markdown as well?


crates/steel-core/src/primitives/strings.rs
```diff
@@ -11,8 +11,6 @@ fn char_upcase(c: char) -> char {
     c.to_ascii_uppercase()
 }
 
-/// # steel/strings
-///
```
jrvidal, line 15: I feel like these titles are unnecessary, since the the doc generator knows the module name already. In the book you can actually see two consecutive`steel/strings` titles.

crates/steel-core/src/primitives/time.rs
```diff
@@ -54,13 +72,16 @@ fn current_milliseconds() -> SteelVal {
...
+///
+/// (current-second) -> int?
 #[function(name = "current-second")]
 fn current_seconds() -> SteelVal {
     use std::time::{SystemTime, UNIX_EPOCH};
 
     match SystemTime::now().duration_since(UNIX_EPOCH) {
         Ok(n) => {
-            let ms = n.as_millis();
+            let ms = n.as_secs();
```
jrvidal, line 84: :warning: I think this was a bug.

crates/steel-core/src/values/json_vals.rs
```diff
@@ -1,43 +1,49 @@
...
+/// (string->jsexpr json) -> any/c?
+///
+/// * json : string?
+///
+/// # Examples
+/// ```scheme
+/// (string->jsexpr "{\"foo\": [3]}") ;; => '#hash((foo . (3)))
+/// ```
+#[function(name = "string->jsexpr")]
+pub fn string_to_jsexpr(value: &SteelString) -> Result<SteelVal> {
```
jrvidal, line 24: I've changed this to use the derive macro, not sure if it has any undesired consequences.

crates/steel-doc/src/main.rs
```diff
@@ -53,7 +53,8 @@ fn main() {
 
                 match value {
                     steel::steel_vm::builtin::Documentation::Markdown(m) => {
-                        writeln!(&mut module_file, "### **{}**", name).unwrap();
+                        let escaped = name.replace("*", "\\*");
+                        writeln!(&mut module_file, "### **{}**", escaped).unwrap();
```
jrvidal, line 57: In lieu of proper markdown generation, this prevents the `*` operator from appearing as `*****` :sweat_smile:

crates/steel-core/src/primitives/time.rs
mattwparas: Good catch!

crates/steel-core/src/values/json_vals.rs
mattwparas: This should be fine, the derive macro is in general the preferred way of implementing it anyway

mattwparas:
To answer your question - Yes if you could run the docgen task and commit that would be great!

crates/steel-doc/src/main.rs
jrvidal: Yes, that's the only one I noticed.

mattwparas (commented):
Overall LGTM - if you wanted to make the change where things prefixed with `#%` don't appear, that would be fine as well - otherwise I can approve as is.

mattwparas (approved):

PR #207, opened 2024-05-11
wrap env::set_var in a closure

RalfJung:
This is in preparation for https://github.com/rust-lang/rust/pull/124636, where we're making `set_var` unsafe. On old (pre-2024) editions, the unsafety error is downgraded to a warning so old code largely keeps working -- except when `set_var` is coerced to a function pointer, as is done by this crate (and by no other crate it seems); then the true type of `set_var` being an `unsafe fn` shines through and the build [fails](https://crater-reports.s3.amazonaws.com/pr-124636/try%23f7e82692c94525c7ff9c5e366c953cc2ea3ca6fc/gh/mattwparas.rucket/log.txt).

So to keep this crate building with future versions of rustc, wrap `set_var` in a closure such that the closure gets coerced to a function pointer, and `set_var` just gets called the normal way.


mattwparas:
Thank you!

mattwparas (approved):

RalfJung:
Thanks for the quick merge :)

PR #209, opened 2024-05-16
improve string->number

jrvidal:
Some minor improvements to `string->number`

## TODO
* [x] Some tests

### DIFF
--- a/crates/steel-core/src/primitives/strings.rs
+++ b/crates/steel-core/src/primitives/strings.rs
@@ -122,7 +122,7 @@ fn string_to_number_impl(value: &str, radix: Option<u32>) -> Result<SteelVal> {
         Some("#x") | Some("#d") | Some("#o") | Some("#b")
     );
 
-    let implicit_radix = radix.filter(|_| explicit_radix);
+    let implicit_radix = radix.filter(|_| !explicit_radix);
 
     let number = expr.into_iter().next().unwrap();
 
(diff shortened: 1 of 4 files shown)


crates/steel-core/src/primitives/strings.rs
```diff
@@ -112,31 +114,44 @@ fn string_to_number_impl(value: &str, radix: Option<u32>) -> Result<SteelVal> {
...
+        },
+        (val @ SteelVal::IntV(_), None) => Ok(val),
+        (SteelVal::BigNum(_), Some(radix)) => match BigInt::from_str_radix(value, radix) {
+            Ok(parsed) => Ok(SteelVal::BigNum(Gc::new(parsed))),
+            Err(_) => Ok(SteelVal::BoolV(false)),
+        },
+
+        (val @ SteelVal::BigNum(_), None) => Ok(val),
+        (svalue @ SteelVal::NumV(_), _) => Ok(svalue),
         _ => Ok(SteelVal::BoolV(false)),
```
jrvidal, line 144: I think I could try to handle all types of numbers, but I might need to expose some of the internals of number tokenization to avoid more duplication.

crates/steel-core/src/primitives/strings.rs
```diff
@@ -112,31 +114,44 @@ fn string_to_number_impl(value: &str, radix: Option<u32>) -> Result<SteelVal> {
...
     }
 }
 
-/// Converts the given string to a number
+/// Converts the given string to a number, with an optional radix.
+/// On failure, it returns `f`
```
mattwparas, line 149:
```suggestion
/// On failure, it returns `#f`
```

mattwparas (approved):

PR #210, opened 2024-05-16
minor repl improvements

jrvidal:
* `Ctrl+C` does not exit the REPL.
* `$1` is bound to the last evaluated expression.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ path = "src/main.rs"
 # This has to line up with the workspace version above
 steel-core = { path = "./crates/steel-core", version = "0.6.0", features = ["dylibs", "markdown", "stacker", "dylib-build"] }
 
+[features]
+interrupt = ["steel-core/interrupt"]
+
 [dependencies]
 once_cell = "1.17.0"
 steel-core = { workspace = true }
--- a/crates/steel-repl/Cargo.toml
+++ b/crates/steel-repl/Cargo.toml
@@ -16,3 +16,4 @@ colored = "2.0.0"
 steel-core = { workspace = true }
 steel-parser = { path = "../steel-parser", version = "0.6.0"}
 dirs = "5.0.1"
+ctrlc = "3.4.4"
--- a/crates/steel-repl/src/repl.rs
+++ b/crates/steel-repl/src/repl.rs
@@ -97,30 +97,36 @@ fn finish_load_or_interrupt(vm: &mut Engine, exprs: String, path: PathBuf) {
     }
 }
 
-fn finish_or_interrupt(vm: &mut Engine, line: String, print_time: bool) {
-    let now = Instant::now();
+fn finish_or_interrupt(vm: &mut Engine, line: String) {
+    let values = match vm.compile_and_run_raw_program(line) {
+        Ok(values) => values,
+        Err(error) => {
+            vm.raise_error(error);
 
-    let res = vm.compile_and_run_raw_program(line);
+            return;
+        }
+    };
 
-    match res {
-        Ok(r) => r.into_iter().for_each(|x| match x {
+    let len = values.len();
+
+    for (i, value) in values.into_iter().enumerate() {
+        let last = i == len - 1;
+
+        if last {
+            vm.register_value("$1", value.clone());
+        }
+
+        match value {
             SteelVal::Void => {}
             SteelVal::StringV(s) => {
                 println!("{} {:?}", "=>".bright_blue().bold(), s);
             }
             _ => {
                 print!("{} ", "=>".bright_blue().bold());
-                vm.call_function_by_name_with_args("displayln", vec![x])
+                vm.call_function_by_name_with_args("displayln", vec![value])
                     .unwrap();
             }
-        }),
-        Err(e) => {
-            vm.raise_error(e);
         }
-    }
-
-    if print_time {
-        println!("Time taken: {:?}", now.elapsed());
     }
 }
 
@@ -233,13 +239,19 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
                     }
                     _ => {
                         // TODO also include this for loading files
-                        finish_or_interrupt(&mut engine.borrow_mut(), line, print_time);
+                        let now = Instant::now();
+
+                        finish_or_interrupt(&mut engine.borrow_mut(), line);
+
+                        if print_time {
+                            println!("Time taken: {:?}", now.elapsed());
+                        }
                     }
                 }
             }
             Err(ReadlineError::Interrupted) => {
                 println!("CTRL-C");
-                break;
+                continue;
             }
             Err(ReadlineError::Eof) => {
                 println!("CTRL-D");
--- a/crates/steel-repl/src/repl.rs
+++ b/crates/steel-repl/src/repl.rs
@@ -1,6 +1,8 @@
 extern crate rustyline;
 use colored::*;
 
+use std::sync::atomic::AtomicBool;
+use std::sync::Arc;
 use std::{cell::RefCell, rc::Rc, sync::mpsc::channel};
 
 use rustyline::error::ReadlineError;
@@ -163,7 +165,10 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
         tx.lock().unwrap().send(()).unwrap();
     };
 
+    let interrupted = Arc::new(AtomicBool::new(false));
+
     vm.register_fn("quit", cancellation_function);
+    vm.with_interrupted(interrupted.clone());
 
     let engine = Rc::new(RefCell::new(vm));
     rl.set_helper(Some(RustylineHelper::new(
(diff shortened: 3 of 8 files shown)


crates/steel-repl/src/repl.rs
```diff
@@ -97,31 +97,37 @@ fn finish_load_or_interrupt(vm: &mut Engine, exprs: String, path: PathBuf) {
...
 
-    match res {
-        Ok(r) => r.into_iter().for_each(|x| match x {
+    let len = values.len();
+
+    for (i, value) in values.into_iter().enumerate() {
+        let last = i == len - 1;
+
+        if last {
+            vm.register_value("$1", value.clone());
```
jrvidal, line 116: Ideally, a `(define $1 "foo")` would shadow this "magic" binding forever, but I imagine that would require deeper changes to the VM/engine.

mattwparas: That should be the behavior as is with `register_value`, so we're good here

jrvidal: FWIW I meant more like:

```scheme
> "foobar"
"foobar"
> $1
=> "foobar"
> (define $1 'xyz)
> "potato"
=> "potato"
> $1  ;; after a explicit `define`, the "magic binding" is gone
=> 'xyz
```

mattwparas: I see, let me take a look and evaluate the difficulty of adding that

crates/steel-repl/src/repl.rs
```diff
@@ -233,13 +239,19 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
...
+                        if print_time {
+                            println!("Time taken: {:?}", now.elapsed());
+                        }
                     }
                 }
             }
             Err(ReadlineError::Interrupted) => {
                 println!("CTRL-C");
-                break;
+                continue;
```
mattwparas, line 254: is the goal here to actually interrupt evaluation? I think if you do something like this:

```scheme
λ > (define (loop) (loop))
λ > (loop)
^C
➜  steel git:(repl)
```

You'll find that it exits the evaluation.

If you do `CTRL-C` while at the top level though, it won't exit though. Was that the goal? If you want it to interrupt evaluation and keep going, I think the changes will be slightly more involved - you'll have to register a handler for ctrl-c and catch it, and just don't exit and continue

jrvidal: Oh, I foolishly assumed signals would _always_ be caught by readline. I'll look into adding a signal handler.

jrvidal: Oh, wait, you'd need a way to _interrupt_ the VM execution. I see you had something like that a while ago (`vm.on_progress`). I think I could hack something together that works, but I worry that I won't be able to gauge whether a naive `break` in some instruction loop might leave the VM in an inconsistent state.

I'm happy to go either way: give it a try or leave `Ctrl-C` as is for now.

mattwparas: I removed `on_progress` simply because I didn't dedicate enough time to see to how it would affect performance in the general case. If we did some benchmarking, I'd be happy to add it back in (perhaps behind a feature flag) - that way we also open the door for debuggers and other kinds of things

the easiest way to do this I think is to special case interruption, rather than use the general purpose `on_progress` - and just have the interpreter loop be a `while <not interrupted>` - and then in the interrupted case, return and error - the error should force all of the clean up that needs to be run.

There might be some edge cases with this, but that would be a nice direction to start if you're interested in contributing that

mattwparas (commented):
This is great - would you be able to run some benchmarks and see if there is any performance change due to this? I'd expect negligible but perhaps worth checking

Also unfortunately I'm gonna be unavailable for the next week and a half or so - so we won't be able to get this merged until I get back.

jrvidal:
> This is great - would you be able to run some benchmarks and see if there is any performance change due to this? I'd expect negligible but perhaps worth checking

```scheme
(define (loop n) (if (= n 0) #f (loop (- n 1))))

(define (try-loop) (loop 100000000))
```

This loop takes close to ~7s on my machine, and there's a reliable difference after this change, around 1%-2%. Not sure how acceptable that is.

jrvidal:
I'm thinking now: perhaps an obvious optimization is to avoid checking the atomic on _every_ instruction, and instead just do it every 1024 instructions or so. I presume the "intermediate counter" can be better optimized to a register.

mattwparas:
It's possible the additional counter and check for every 1024 could be more expensive than checking the atomic I suppose - I'm not opposed to either, whichever one is faster! 

Also we can probably put this behind a feature flag before rolling it out everywhere. I could imagine that you could have use cases where you don't need to interrupt it and shouldn't have to pay the cost (even though the cost is minimal)

jrvidal:
> It's possible the additional counter and check for every 1024 could be more expensive than checking the atomic I suppose

^ My benchmarks seem to indicate that you were right... CPUs are complicated beasts :shrug: 

> we can probably put this behind a feature flag

I've made it optional, both at runtime and compile time. Wrapping it in an `Option` penalizes the REPL a bit more in favor of speeding up running scripts, which I think it makes sense.

crates/steel-repl/Cargo.toml
```diff
@@ -16,3 +16,4 @@ colored = "2.0.0"
 steel-core = { workspace = true }
 steel-parser = { path = "../steel-parser", version = "0.6.0"}
 dirs = "5.0.1"
+ctrlc = "3.4.4"
```
mattwparas, line 19: Not directly related to this line change, but I think in this TOML file you'll have to change the workspace version for steel to have the `interrupt` feature flag

mattwparas (commented):
Overall the changes look okay (just left a small comment), although I'd like to align with you on the benchmarks you're using so we can maybe have a standard set to run moving forward. There are benchmarks that exist, although I would like to consolidate them into a one liner to run everything to make it easy for everyone - then we can get a dump of the report pre and post change. I can work on that this week if that is an alright timeline for you?

crates/steel-repl/Cargo.toml
jrvidal: Hm, I'm not sure what you mean. I did forget to expose the feature at the "top level" package (which I've now changed). I think that should be enough?

Cargo.toml
```diff
@@ -21,6 +21,9 @@ path = "src/main.rs"
 # This has to line up with the workspace version above
 steel-core = { path = "./crates/steel-core", version = "0.6.0", features = ["dylibs", "markdown", "stacker", "dylib-build"] }
```
mattwparas, line 22: I figured you would need to add it here, so that the application uses it

crates/steel-repl/Cargo.toml
mattwparas: See this comment: https://github.com/mattwparas/steel/pull/210/files#r1625130105

Cargo.toml
jrvidal: Ah, it was unclear to me whether you wanted this enabled by default or not, that's why I exposed it as a top-level feature for the main binary. If you want it enabled by default, maybe it would make sense to add it to `default-features` so that consumers can still disable it?

mattwparas: Ah right, that makes sense, thanks for the clarification. For now lets leave it as is, and once we're done with the benchmarks we can decide if its default or not

mattwparas:
> > if that is an alright timeline for you?
> 
> Sure, whatever works for you 👍

So I didn't get around to the benchmarks - did you want to merge this as is? Since its opt in, I could live on it for a bit and run some benchmarks separately and see what the overall cost is. I'd imagine about 1% is probably right, which I think I could live with.

jrvidal:
> did you want to merge this as is (including the $1 stuff)?

I'm in no particular hurry to merge anything, please do what you think it's best. I guess, worse case scenario, if you find the benchmarks to be less than ideal, we could scrap the interrupt stuff out or rethink it.

mattwparas:
Looks like I'm getting about 1-3% performance decrease, however given that its optional I'm inclined to include this change since I think being able to interrupt and continue working at the repl is most likely worth it.

jrvidal:
The numbers I'm getting are kind of all over the place:

```
ackermann-3-3                   =>  +7.7%
binary-trees                    =>  +1.3%
engine-creation                 =>  +0.7%
fib-28                          =>  -3.4%
filter-big                      =>    +5%
map-big                         =>  +7.3%
multiple-transducers            =>  -4.1%
range-big                       => +24.6%
register-fn                     =>  -6.5%
ten-thousand-iterations         =>  +9.7%
ten-thousand-iterations-letrec  =>  +6.1%
thread-creation                 =>  +0.1%
transducer-map                  =>  +1.9%
trie-sort-without-optimizations =>  -0.5%

```

This is the diff between the estimated mean for each benchmark, in alphabetic order. Just in case, did you remember to enable the `interrupt` feature when testing this branch?

mattwparas:
> The numbers I'm getting are kind of all over the place:
> 
> ```
> ackermann-3-3                   =>  +7.7%
> binary-trees                    =>  +1.3%
> engine-creation                 =>  +0.7%
> fib-28                          =>  -3.4%
> filter-big                      =>    +5%
> map-big                         =>  +7.3%
> multiple-transducers            =>  -4.1%
> range-big                       => +24.6%
> register-fn                     =>  -6.5%
> ten-thousand-iterations         =>  +9.7%
> ten-thousand-iterations-letrec  =>  +6.1%
> thread-creation                 =>  +0.1%
> transducer-map                  =>  +1.9%
> trie-sort-without-optimizations =>  -0.5%
> ```
> 
> This is the diff between the estimated mean for each benchmark, in alphabetic order. Just in case, did you remember to enable the `interrupt` feature when testing this branch?

I did enable the interrupt feature, but yeah those are really all over the place. Maybe some of these benchmarks are too noisy?

mattwparas (approved):
Lets go ahead and merge this if thats okay - we can keep iterating on the implementation since I want the interrupt feature myself

PR #211, opened 2024-05-17
nix: more granular source files

evanrichter:
this prevents rebuilding steel when changing unrelated files, such as flake.nix itself

gitignore.nix was removed in favor of the new fileset api in nixpkgs


mattwparas:
Thanks!

PR #217, opened 2024-06-12
support improper lists

jrvidal:
This is a somewhat incomplete PR to support improper lists at the syntax level.

It is incomplete b/c macro patterns and matching has not been updated, so things like the macro definition of `case-lambda` does not work properly. However, it doesn't work properly at the moment either, so it might make sense to merge without fully solving the macro situation.

### TODO
* [x] Add more tests (but first let's see if we actually want to merge it)


crates/steel-parser/src/lexer.rs
```diff
@@ -271,9 +271,9 @@ impl<'a> Lexer<'a> {
     fn read_word(&mut self) -> TokenType<&'a str> {
         while let Some(&c) = self.chars.peek() {
             match c {
-                '(' | '[' | ')' | ']' => break,
+                '(' | '[' | ')' | ']' | '{' | '}' => break,
                 c if c.is_whitespace() => break,
-                '\'' => {
+                '\'' | '"' | '`' | ';' | ',' => {
```
jrvidal, line 276: This is the original rabbit hole I was getting into: improving how the reader handles identifiers, in preparation for supporting escaped identifiers `|...|`. Alas, more work for the future.

crates/steel-parser/src/lexer.rs
```diff
@@ -308,6 +309,7 @@ impl<'a> Lexer<'a> {
             NEG_INFINITY => TokenType::Number(RealLiteral::Float(f64::NEG_INFINITY).into()),
             NAN => TokenType::Number(RealLiteral::Float(f64::NAN).into()),
             NEG_NAN => TokenType::Number(RealLiteral::Float(f64::NAN).into()),
+            "|.|" => TokenType::Identifier("."),
```
jrvidal, line 312: I'm hardcoding this escaped identifier for now so I can fix `stack_struct.scm`

steel-examples/stack_struct.scm
```diff
@@ -57,11 +57,11 @@
...
 (define-syntax impl
   (syntax-rules ()
+    [(impl struct-name (define/method (a this b ...) body ...))
+     (def-method struct-name (define/method (a this b ...) body ...))]
     [(impl struct-name (define/method (a this b ...) body ...) c ...)
      (begin (def-method struct-name (define/method (a this b ...) body ...))
-            (impl struct-name c ...))]
-    [(impl struct-name (define/method (a this b ...) body ...))
-     (def-method struct-name (define/method (a this b ...) body ...))]))
+            (impl struct-name c ...))]))
```
jrvidal, line 64: I think these needed to be swapped to work properly.

crates/steel-core/src/rvals.rs
```diff
@@ -836,6 +836,7 @@ pub enum SerializableSteelVal {
     MutFunc(MutFunctionSignature),
     HashMapV(Vec<(SerializableSteelVal, SerializableSteelVal)>),
     ListV(Vec<SerializableSteelVal>),
+    Pair(Box<(SerializableSteelVal, SerializableSteelVal)>),
```
jrvidal, line 839: I had to add this new variant because of some [failing test](https://github.com/mattwparas/steel/actions/runs/9490147478/job/26152981083). I can't reproduce locally though :thinking:

mattwparas:
This is great, thank you! I'll take a look this evening. Adjusting macros to include this won't be _too_ hard but we don't need to have that as part of this change since it will get more complicated pretty quickly.

crates/steel-parser/src/ast.rs
```diff
@@ -977,6 +977,8 @@ pub struct List {
 impl PartialEq for List {
     fn eq(&self, other: &Self) -> bool {
         self.args == other.args
+            && self.improper == other.improper
+            && self.location == other.location
```
jrvidal, line 981: I'm making list less special, and instead rely on an `Eraser` visitor to help in tests.

crates/steel-parser/src/visitors.rs
```diff
@@ -0,0 +1,300 @@
+use crate::ast::*;
```
jrvidal, line 1: Moved from `steel-core`, to be able to create `Eraser`.

crates/steel-parser/src/parser.rs
```diff
@@ -1350,79 +1380,100 @@ impl ASTLowerPass {
                             let value = std::mem::replace(value, List::new(vec![]));
 
                             *expr = match &a.syn.ty {
-                                TokenType::If => parse_if(value.args.into_iter(), a.syn),
+                                TokenType::If => {
+                                    parse_if(value.args_proper(TokenType::If)?.into_iter(), a.syn)
+                                }
```
jrvidal, line 1385: These help with things like `(let ((x 1)) . 1)`. I'm not sure this properly catches all possible mistakes, since lowering is a bit all over the place.

(Also, I'm assuming that no special form allows dots, but I haven't checked)

mattwparas: The lowering is all over the place for a few reasons, and in general I'd like to clean that up - but the reason was that originally I parsed directly into the core forms. Once macros were added, I instead parsed into lists before lowering to the core forms - and also when dealing with kernel level macros, you have to go back and forth sometimes.

I think this is fine though - I will take a holistic pass and see where all the lowering is happening

crates/steel-parser/src/lexer.rs
mattwparas: For what its worth we could change `stack_struct.scm` - its just a test I've had around for a long time. I can see how it behaves on racket before we do anything though

crates/steel-core/src/rvals.rs
mattwparas: Interesting that you weren't able to reproduce locally, but I think this change is correct. When creating a new thread we have to serialize the value since each thread manages its own memory since nothing is really thread safe at the moment.

crates/steel-parser/src/lexer.rs
mattwparas: So are `{` and `}` no longer legal in identifiers?

mattwparas (commented):
Overall lgtm - just had a couple questions / comments. Once those are resolved we can merge

crates/steel-parser/src/lexer.rs
jrvidal: They're not, which is also in line with Racket.

crates/steel-core/src/rvals.rs
jrvidal: I just tried to reproduce in my fork using the same GH actions and nothing... :thinking:

mattwparas (approved):

PR #223, opened 2024-06-20
better matching for parens

jrvidal:
* Different types of parenthesis match each other e.g. `(a 1]` is forbidden:
  ```
  λ > (a]
  error[E09]: Parse
    ┌─ :1:3
    │
  1 │ (a]
    │   ^ Parse: Unexpected token: CloseParen(Square)
  ```
* Refactored paren highlighting in the REPL.
* Validator in the REPL is more precise now. For instance, `( "asdf)" ` was not recognized as an incomplete expression.

### DIFF
--- a/crates/steel-repl/src/highlight.rs
+++ b/crates/steel-repl/src/highlight.rs
@@ -4,9 +4,7 @@ use colored::*;
 use std::{cell::RefCell, rc::Rc};
 
 use rustyline::highlight::Highlighter;
-use rustyline::validate::{
-    MatchingBracketValidator, ValidationContext, ValidationResult, Validator,
-};
+use rustyline::validate::{ValidationContext, ValidationResult, Validator};
 
 use rustyline::{hint::Hinter, Context};
 use rustyline_derive::Helper;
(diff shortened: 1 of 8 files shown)


crates/steel-repl/src/highlight.rs
```diff
@@ -44,11 +39,36 @@ impl RustylineHelper {
...
+                }
+                _ => {}
+            }
+        }
+
+        if balance > 0 || unfinished_string {
+            Ok(ValidationResult::Incomplete)
+        } else {
+            Ok(ValidationResult::Valid(None))
+        }
```
jrvidal, line 71: I'm not a big fan of returning `ValidationResult::Invalid`. The message appears right next to the input and it's not very user-friendly IMO:
```
λ > (aadf]Mismatched brackets: '(' is not properly closed
```

So the case of `balance < 0` or mismatched parentheses is left to the parser to throw a much more precise error.

crates/steel-repl/src/highlight.rs
```diff
@@ -75,27 +93,66 @@ impl Highlighter for RustylineHelper {
...
 
             // if token.span().end() > pos {
             //     return self.highlighter.highlight(line, pos);
             // }
 
             match token.typ() {
-                // steel::parser::tokens::TokenType::OpenParen => todo!(),
-                // steel::parser::tokens::TokenType::CloseParen => todo!(),
+                TokenType::OpenParen(paren) if paren_to_highlight.is_none() => {
+                    if token.span.start == pos || (token.span.start == pos + 1 && cursor.is_none())
```
jrvidal, line 114: We highlight the match of:
* The parenthesis right under the cursor if any, or
* The _closing_ parenthesis right to the _left_ of the cursor, if any, or
* The _opening_ parenthesis right to the _right_ of the cursor, if any.

The last point is a deviation from current behavior, where we always peek to the left, regardless of closing/opening. I find this slightly more... symmetrical? But I can easily revert to the current behavior.

mattwparas: I think this is reasonable!

mattwparas (approved):
Thanks! This is great

PR #224, opened 2024-06-21
fix recyclable drop

jrvidal:
Fixes a panic I noticed while trying the helix integration. My guess is that:
1. Static storage transitively contains recyclable values.
2. The storage gets dropped at thread exit.
3. Some recyclable value `Drop` impl is called and attempts to access the thread local.
4. And that panics b/c the thread local `Drop` call is ongoing.

### DIFF
--- a/crates/steel-core/src/values/recycler.rs
+++ b/crates/steel-core/src/values/recycler.rs
@@ -72,7 +72,7 @@ macro_rules! impl_recyclable {
 
         impl Recyclable for $t {
             fn put(mut self) {
-                $tl.with(|p| {
+                let _ = $tl.try_with(|p| {
                     let p = p.try_borrow_mut();
 
                     if let Ok(mut p) = p {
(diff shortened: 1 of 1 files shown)


mattwparas:
Did this happen at exit? I must have not run into it myself

Anyway this look good to me

mattwparas (approved):

jrvidal:
> Did this happen at exit? I must have not run into it myself

Yes, it always happened when exiting helix. FWIW I ran your `helix` branch together with the latest `master` from Steel (with some minor modifications to get it to compile).

mattwparas:
I must just be behind then. I'll take a look this weekend

mattwparas:
@jrvidal What were the other issues you had with helix + steel? I'm trying with latest master right now and running into some unexpected behavior with multi arity functions from the built ins

jrvidal:
I just uploaded a couple branches:

https://github.com/jrvidal/helix/tree/mwp-steel-integration-hack
https://github.com/jrvidal/steel/tree/steel-for-helix

I might be able to add more details later.

PR #227, opened 2024-06-27
more methods: ports, chars

jrvidal:
This started as "let's add some missing R<sup>7</sup>RS stuff" and ended up a little all over the place :sweat_smile: 

## TODO
* [x] `xtask docgen`


crates/steel-core/src/primitives/bytevectors.rs
```diff
@@ -206,19 +206,6 @@ pub fn bytes_length(value: &SteelByteVector) -> usize {
...
 }
 
-/// Converts the given string to a bytevector
-///
-/// # Examples
-/// ```scheme
-/// (string->bytes "Apple") ;; => (bytes 65 112 112 108 101)
-/// ```
-#[function(name = "string->bytes")]
-pub fn string_to_bytes(value: &SteelString) -> Result<SteelVal> {
```
jrvidal, line 216: :information_source: Moved to `strings.rs`

crates/steel-core/src/primitives/bytevectors.rs
```diff
@@ -320,3 +307,51 @@ pub fn bytes_append(value: &SteelByteVector, other: &SteelByteVector) -> Result<
...
+///
+/// * buf : bytes?
+/// * start: int? = 0
+/// * end: int? = (bytes-length buf)
+///
+/// # Examples
+/// ```scheme
+/// (utf8->string (bytes #xe5 #x8d #x83 #xe8 #x91 #x89)) ;; => "千葉"
+/// ```
+#[function(name = "utf8->string")]
```
jrvidal, line 323: :question: Should we call this `bytes->string` instead?

crates/steel-core/src/primitives/ports.rs
```diff
@@ -1,13 +1,17 @@
...
 use crate::values::port::new_rc_ref_cell;
 use crate::values::port::{SteelPort, SteelPortRepr};
+use crate::values::structs::{make_struct_singleton, StructTypeDescriptor};
 
+use once_cell::unsync::Lazy;
 use steel_derive::function;
 
 thread_local! {
-    pub static EOF_OBJECT: SteelString = "eof".into();
+    static EOF_OBJECT: Lazy<(SteelVal, StructTypeDescriptor)>= Lazy::new(|| {
```
jrvidal, line 12: :information_source: Added a proper EOF object, as a singleton of a unique struct.

crates/steel-core/src/primitives/strings.rs
```diff
@@ -701,6 +735,152 @@ pub fn string_append(mut rest: RestArgsIter<'_, &SteelString>) -> Result<SteelVa
...
+    }
+
+    let range = bounds(value.as_str(), start, end, "string->utf8")?;
+
+    let chars: im_rc::Vector<_> = value[range].chars().map(SteelVal::CharV).collect();
+
+    Ok(SteelVal::VectorV(Gc::new(chars).into()))
+}
+
+fn bounds(
```
jrvidal, line 830: :information_source: Helper to select string ranges.

crates/steel-core/src/values/port.rs
```diff
@@ -49,10 +50,11 @@ pub enum SteelPortRepr {
     FileOutput(String, BufWriter<File>),
     StdInput(Stdin),
     StdOutput(Stdout),
+    StdError(Stderr),
     ChildStdOutput(BufReader<ChildStdout>),
     ChildStdInput(BufWriter<ChildStdin>),
-    StringInput(BufReader<Cursor<Vec<u8>>>),
-    StringOutput(BufWriter<Vec<u8>>),
```
jrvidal, line 55: I don't think the buffered variants are adding anything useful here.

crates/steel-core/src/values/structs.rs
```diff
@@ -526,6 +512,38 @@ pub fn make_struct_type(args: &[SteelVal]) -> Result<SteelVal> {
     ))
 }
 
+pub fn make_struct_singleton(name: &str) -> (SteelVal, StructTypeDescriptor) {
```
jrvidal, line 515: :information_source: Used for the EOF object

crates/steel-core/src/primitives/ports.rs
```diff
@@ -172,9 +210,10 @@ pub fn write_line(port: &SteelPort, line: &SteelVal) -> Result<SteelVal> {
 }
 
 #[function(name = "raw-write")]
-pub fn write(port: &SteelPort, line: &SteelVal) -> Result<SteelVal> {
+pub fn write(line: &SteelVal, rest: RestArgsIter<&SteelPort>) -> Result<SteelVal> {
```
jrvidal, line 213: :warning: I've changed a few signatures to make the port optional.

crates/steel-core/src/primitives/bytevectors.rs
mattwparas: I think the r7rs spec uses `utf8->string`, but racket uses `bytes->string/utf8` (see [here](https://docs.racket-lang.org/reference/bytestrings.html#%28def._%28%28quote._~23~25kernel%29._bytes-~3estring%2Futf-8%29%29) - maybe we just include both?

crates/steel-core/src/primitives/ports.rs
mattwparas: Thanks! I've been meaning to do this and just kept forgetting.

This works fine and I don't have anything against it. Did you consider something like this?:

```rust
struct EofObject;

impl Custom for EofObject {
    // probably implement the equality method?
}
```

Actually having typed that out, I think I like your solution better since we can just inherit struct equality and everything like that

crates/steel-core/src/primitives/ports.rs
```diff
@@ -29,10 +33,21 @@ pub fn port_module() -> BuiltInModule {
...
+        .register_native_fn_definition(READ_BYTE_DEFINITION)
+        .register_native_fn_definition(READ_CHAR_DEFINITION)
+        .register_native_fn_definition(WRITE_BYTE_DEFINITION)
+        .register_native_fn_definition(WRITE_BYTES_DEFINITION)
+        .register_native_fn_definition(PEEK_BYTE_DEFINITION);
     module
 }
 
+// TODO: implement textual-port? and binary-port?
+
```
mattwparas, line 50: Were you planning on doing this in this PR?

crates/steel-core/src/primitives/ports.rs
mattwparas: I think this should be fine since this is intended as an internal API

crates/steel-core/src/primitives/strings.rs
```diff
@@ -183,73 +216,110 @@ pub fn string_constructor(rest: RestArgsIter<'_, char>) -> Result<SteelVal> {
...
-pub fn string_ci_greater_than(left: &SteelString, right: &SteelString) -> bool {
-    left.to_lowercase() > right.to_lowercase()
+// TODO: avoid allocation in `-ci` variants
+
+macro_rules! impl_str_comparison {
+    (-ci, $name:ident, $ext_name:literal, $mode:literal, $op:expr) => {
+        #[doc = concat!("Compares strings lexicographically (as in\"", $mode, "\"),")]
+        #[doc = "in a case insensitive fashion."]
+        #[doc = ""]
+        #[doc = concat!("(", $ext_name, " s1 s2 ... ) -> bool?")]
```
mattwparas, line 226: Nice, I like this!

mattwparas (commented):
Just a question, otherwise I think this is good to go!

crates/steel-core/src/primitives/ports.rs
jrvidal: Would it make sense to rename them to `#%raw-write`, etc.?

crates/steel-core/src/primitives/ports.rs
jrvidal: No, I took the liberty to leave a couple of TODOs.

There are a few things that I believe would require a deeper refactor of ports, so I stopped before this PR got too unwieldy. FWIW, I was thinking about:

* Textual/binary split
* `peek-char`
* Bidirectional ports (that are both input and output and can be closed independently).

crates/steel-core/src/values/port.rs
```diff
@@ -296,42 +333,56 @@ impl SteelPortRepr {
             SteelPortRepr::FileOutput(_, _)
                 | SteelPortRepr::StdOutput(_)
                 | SteelPortRepr::DynWriter(_)
-        )
-    }
-
-    pub fn is_textual(&self) -> bool {
```
jrvidal, line 302: :warning: I deleted this since it is currently unused and not quite accurate.

jrvidal:
> Just a question, otherwise I think this is good to go!

Cool, let me run `docgen` before merging (plus maybe take a last look/add some more tests)

crates/steel-core/src/primitives/ports.rs
mattwparas: probably would make more sense to do that if you were willing to include that change here, that was it is more clear that it is supposed to be largely private

crates/steel-core/src/primitives/ports.rs
mattwparas: Got it - is the implementation then just fundamentally incompatible with `peek-char`? I haven't looked at ports in a bit so I'd have to refresh my memory

jrvidal: (I also remembered another one: `char-ready?`, `u8-ready?`)

re: `peek-char`: I actually added some context [here](https://github.com/mattwparas/steel/pull/227/files#diff-0102212dc5c4ef8c26b71d2b0b861ad2ef745eb95081fed3b29b54f95f68bb6bR270-R272).

TBF I'm not sure how fundamental the changes would need to be, I just went for quick wins and avoided more convoluted work for this PR.

crates/steel-derive/src/lib.rs
```diff
@@ -39,26 +39,37 @@ fn parse_key_value_pairs(args: &Punctuated<Meta, Token![,]>) -> HashMap<String,
...
+                    Expr::Lit(ExprLit {
+                        lit: Lit::Bool(b), ..
+                    }) => {
+                        map.insert(key, b.value().to_string());
+                        break;
+                    }
+                    Expr::Group(ExprGroup { expr, .. }) => {
+                        value = &**expr;
+                    }
+                    _ => {}
```
mattwparas, line 61: Sorry, just noticing this - will this loop infinitely if there isn't a match? e.g. you don't use either a literal or a group?

jrvidal: Oops, good catch

mattwparas (approved):
Thank you!

PR #228, opened 2024-06-30
Modern metadata for Nix

bddvlpr:
Currently, in order to detect a binary or entry point to a program in Nix, you can use lib.getExe which requires the meta.mainProgram variable to be set. If not, you can fallback to lib.getExe'. Using this function allows you to set the mainProgram yourself which I'm not a fan of as it could result in it being changed over time. Also added other metadata that packages usually have to be complete :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -55,6 +55,12 @@
             popd
             rm "$out/bin/cargo-steel-lib"
           '';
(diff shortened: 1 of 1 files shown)


mattwparas:
Thanks!

PR #230, opened 2024-07-03
add macro tests

jrvidal:
This adds a minor `syntax-tests.scm` test file, together with minor additions to support it. I think many of them make sense even if the main idea of this PR is unwelcome.

## TODO
* [ ] Failed tests

### DIFF
--- a/cogs/syntax-tests.scm
+++ b/cogs/syntax-tests.scm
@@ -1,34 +1,16 @@
-(require-builtin "steel/process")
-(require-builtin "steel/transducers")
+(require-builtin steel/process)
+(require-builtin steel/transducers)
+(require-builtin steel/meta)
 
 (require "tests/unit-test.scm"
          (for-syntax "tests/unit-test.scm"))
 
-(define (eval-in-child datums)
-  (define out-path "_syntax_tmp_.scm")
-  (define out (open-output-file out-path))
-
-  (transduce datums (into-for-each (lambda (datum)
-                                           (write datum out)
-                                           (newline out))))
-  (define interpreter (car (command-line)))
-  (define cmd (command interpreter (list out-path)))
-  (set-piped-stdout! cmd)
-  (define child (Ok->value (spawn-process cmd)))
-  (define child-errors (child-stderr child))
-  (define exit-result (wait child))
-  (define stderr (read-port-to-string child-errors))
-  (delete-file! out-path)
-
-  (cons (Ok->value exit-result) stderr))
-
 (define (check-syntax-error? name input expected . rest)
   (define (impl name input expected)
-      (define output (eval-in-child input))
-      (define exit-code-assert (if (> (car output) 0) #t `(exit= ,(car output))))
-      (define message-assert (if (string-contains? (cdr output) expected) #t `(message= ,(cdr output))))
-      (check-equal? (string-join `(,name ", compilation fails")) exit-code-assert #t)
-      (check-equal? (string-join `(,name ", message")) message-assert #t))
+      (define error-message
+        (~> (run! (Engine::new) input) Err->value error-object-message))
+      (define message-assert (or (string-contains? error-message expected) `(message= ,error-message expected= ,expected)))
+      (check-equal? (string-join `(,name " [compilation fails]"))  message-assert #t))
   (if (eq? name 'skip)
       (skip-compile #f)
       (impl name input expected)))
--- a/cogs/syntax-tests.scm
+++ b/cogs/syntax-tests.scm
@@ -1,0 +1,103 @@
+(require-builtin "steel/process")
+(require-builtin "steel/transducers")
+
+(require "tests/unit-test.scm"
+         (for-syntax "tests/unit-test.scm"))
+
+(define (eval-in-child datums)
+  (define out-path "_syntax_tmp_.scm")
+  (define out (open-output-file out-path))
+
+  (transduce datums (into-for-each (lambda (datum)
+                                           (write datum out)
+                                           (newline out))))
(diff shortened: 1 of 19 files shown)


cogs/syntax-tests.scm
```diff
@@ -0,0 +1,103 @@
+(require-builtin "steel/process")
+(require-builtin "steel/transducers")
+
+(require "tests/unit-test.scm"
+         (for-syntax "tests/unit-test.scm"))
+
+(define (eval-in-child datums)
```
jrvidal, line 7: :information_source: I want to test expected failures when compiling macros. I don't think we have `eval` or something similar, so I've come up with a gimmick: just write the code to some file and run it in a separate process.

crates/steel-core/src/primitives/process.rs
```diff
@@ -98,16 +113,21 @@ impl ChildProcess {
...
+            .map_err(SteelErr::from)?;
+
+        match exit_status.code() {
+            Some(code) => Ok(code.into()),
+            None => Ok(false.into()),
+        }
     }
 
-    pub fn wait(&mut self) -> SteelResult<ProcessExitStatus, SteelErr> {
+    pub fn wait(&mut self) -> SteelResult<SteelVal, SteelErr> {
```
jrvidal, line 130: :warning: I've changed this signature to return the exit code or `#f`.

crates/steel-core/src/rerrs.rs
```diff
@@ -240,7 +241,13 @@ impl SteelErr {
     pub fn emit_result(&self, file_name: &str, file_content: &str) {
         // let opts = Opts::();
         // let config = codespan_reporting::term::Config::default();
-        let writer = StandardStream::stderr(ColorChoice::Always);
+        let color = if std::io::stderr().is_terminal() {
+            ColorChoice::Auto
+        } else {
+            ColorChoice::Never
+        };
```
jrvidal, line 248: :information_source:

* Don't colorize output if stderr is not a terminal.
* Use `Auto` instead of `Always` to respect `NO_COLOR` and such.

crates/steel-core/src/rvals/cycles.rs
```diff
@@ -178,7 +178,11 @@ impl CycleDetector {
                 }
             }
             Pair(p) => {
-                write!(f, "({} . {})", p.car(), p.cdr())
+                write!(f, "(")?;
+                self.format_with_cycles(&p.car, f, FormatType::Normal)?;
+                write!(f, " . ")?;
+                self.format_with_cycles(&p.cdr, f, FormatType::Normal)?;
+                write!(f, ")")
```
jrvidal, line 185: Small bugfix

crates/steel-core/src/rvals.rs
```diff
@@ -2124,8 +2124,7 @@ impl fmt::Display for SteelVal {
         // at the top level, print a ' if we are
         // trying to print a symbol or list
         match self {
-            SymbolV(_) | ListV(_) => write!(f, "'")?,
-            VectorV(_) => write!(f, "'#")?,
```
jrvidal, line 2128: :question: This is a change in behavior:

```
;; before
λ > (write 'a)
'a

;; after
λ > (write 'a)
a
```

I think it kinda makes sense for `write` to output something that can be parsed back with `read`. AFAICS that is not a requirement of R7RS, but it is the [documented behavior](https://docs.racket-lang.org/reference/Writing.html#%28def._%28%28quote._~23~25kernel%29._write%29%29) in Racket:

> Writes datum to out, normally in such a way that instances of core datatypes can be read back in

src/lib.rs
```diff
@@ -90,9 +90,7 @@ pub fn run(clap_args: Args) -> Result<(), Box<dyn Error>> {
 
             if let Err(e) = res {
                 e.emit_result(path.to_str().unwrap(), &contents);
-                // process::exit(1);
-
-                return Err(Box::new(e));
```
jrvidal, line 95: :warning: Returning the error here ends up printing duplicated info, that I'm not sure it's desired.

crates/steel-core/src/rvals.rs
jrvidal: Hm, I might have broken some tests because of this :thinking:

mattwparas: So I'll have to look at the tests; its possible that the tests need to be updated since they're either still using the old behavior or were just fundamentally wrong to begin with.

I think that matching behavior with Racket in this case is good - in fact, I'm surprised that I didn't check that it was matching before. In any case I'm okay with this behavior change, assuming we can make sure it doesn't break other things / update everything else to work with it

cogs/syntax-tests.scm
mattwparas: So there is an `eval` - however it basically just works by embedding an engine instance itself, and then we act on that. More or less equivalent to what you've done, but runs in the same process: https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/steel_vm/meta.rs#L225

You also can do something like this, where you pass in a list of expressions:


```
(run! (Engine::new) '((define (foo) 10) (foo))
```

I haven't tested it with macros and making sure all things line up, but it might be easier than creating another process.

crates/steel-core/src/rvals.rs
jrvidal: Just needed to remove the `'` from a few expectations :+1: 

> AFAICS that is not a requirement of R7RS

I just noticed that it is a suggestion in R6RS.

mattwparas (approved):
Thanks for making the adjustments with switching to use the engine API - also apologies for the merge conflicts!

PR #241, opened 2024-07-10
revamp macro matching

jrvidal:
This is a mixed bag of improvements to the expander. My actual end goal was better support for improper lists, but I've accrued too many small changes.

### DIFF
--- a/cogs/syntax-tests.scm
+++ b/cogs/syntax-tests.scm
@@ -22,6 +22,62 @@
                     [(_ a) ])))
   "syntax-rules requires only one pattern to one body")
 
+(check-syntax-error? "repeated pattern variables"
+  '(
+    (define-syntax repeated-vars
+      (syntax-rules ()
+                    [(_ a (b a)) a])))
+  "repeated pattern variable a")
+
+(check-syntax-error? "repeated pattern variables, macro name"
+  '(
+    (define-syntax repeated-vars-macro-name
+      (syntax-rules ()
+                    [(foo a (b foo)) a])))
+  "repeated pattern variable foo")
+
+(check-syntax-error? "multiple ellipsis"
+  '(
+    (define-syntax many-ellipsis
+      (syntax-rules ()
+                    [(_ (a ... b ...)) (a ...)])))
+  "pattern with more than one ellipsis")
+
+(check-syntax-error? "ellipsis in cdr"
+  '(
+    (define-syntax ellipsis-tail
+      (syntax-rules ()
+                    [(_ (a . ...)) (a ...)])))
+  "ellipsis cannot appear as list tail")
+
+(check-syntax-error? "ellipsis in nested cdr"
+  '(
+    (define-syntax ellipsis-tail-nested
+      (syntax-rules ()
+                    [(_ ((a) . ...)) (a ...)])))
+  "ellipsis cannot appear as list tail")
+
+(check-syntax-error? "ellipsis in macro name"
+  '(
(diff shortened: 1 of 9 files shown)


crates/steel-core/src/parser/expander.rs
```diff
@@ -625,251 +662,235 @@ impl MacroPattern {
...
 
-pub fn match_vec_pattern(args: &[MacroPattern], list: &[ExprKind]) -> bool {
-    let mut token_iter = list.iter();
-
-    for pat in args {
-        if let Some(val) = token_iter.next() {
-            // dbg!(&pat);
-            // println!("{}", val);
-
-            match pat {
```
jrvidal, line 658: :information_source: This diff looks big, but this `match` is moved almost verbatim to line 727. This helper is much flatter now.

crates/steel-core/src/parser/expander.rs
```diff
@@ -625,251 +662,235 @@ impl MacroPattern {
...
-                        }
-                    } else if let ExprKind::Quote(_) = val {
-                        // TODO: Come back here
-                        return matches!(vec.as_slice(), &[MacroPattern::Quote(_)]);
-                        // return true;
-                    } else {
-                        // log::debug!("Matching failed - atom does not match list");
+        let Some(val) = exprs_iter.next() else {
+            return false;
+        };
```
jrvidal, line 725: :information_source: Since we handle `Many*` separately, we can now demand `val` to be present.

crates/steel-parser/src/ast.rs
```diff
@@ -1011,7 +1011,22 @@ impl List {
...
+
+        let ExprKind::List(l) = last else {
+            self.args.push(last);
+
+            self.improper = true;
+            return self;
+        };
+
+        self.args.extend(l.args);
+        self.improper = l.improper;
```
jrvidal, line 1029: :information_source: This collapses e.g. `(a b . (c . d))` into `(a b c . d)`, and allows us to write stuff like `(define . (x . (1 . ())))` :upside_down_face:

cogs/syntax-tests.scm
```diff
@@ -22,6 +22,62 @@
                     [(_ a) ])))
   "syntax-rules requires only one pattern to one body")
 
+(check-syntax-error? "repeated pattern variables"
```
jrvidal, line 25: :information_source: All the new stuff should have tests in this file.

mattwparas:
Thanks for this! This is some of the oldest code in steel and its a little gnarly, sorry you had to dig through all of that

Overall the changes LGTM, however I'd like to spend just a bit more time looking at the code locally before we merge. I'll have some time later today that I can do that.

crates/steel-core/src/parser/expander.rs
```diff
@@ -625,251 +677,235 @@ impl MacroPattern {
...
+        patterns = &patterns[1..];
+    }
+
+    let mut exprs_iter = list.iter();
+
+    let len_matches = patterns.len() == list.len() || {
+        let has_ellipsis = patterns.iter().any(|pat| pat.is_many());
+
+        has_ellipsis && list.len() >= (patterns.len() - 1)
+    };
```
jrvidal, line 701: :warning: I just realized that this check is duplicated/redundant with `match_case()`. I have a subsequent PR ready right after this one so, if that's OK, I'd leave this as is for now.

mattwparas (approved):

PR #251, opened 2024-07-19
Shrink AST size, decrease allocated memory while idle

mattwparas:


cogs/syntax-tests.scm
```diff
@@ -173,97 +158,105 @@
 
 (define-syntax lexical-capture
   (syntax-rules ()
-                [(_) bound-x]))
+    [(_) bound-x]))
 
-(let
-  [(bound-x 'inner)]
-  (skip-compile (check-equal? "hygiene, lexical capture" (lexical-capture) 3)))
+(let ([bound-x 'inner]) (check-equal? "hygiene, lexical capture" (lexical-capture) 3))
```
mattwparas, line 163: @jrvidal - lexical capture should be fixed with this PR once it is ready and cleaned up.

I think a good chunk of this will need to be reworked in order to support local syntax-rules and let-syntax, but overall I understand the macro expansion a bit more than previously.

crates/steel-core/src/parser/parser.rs
```diff
@@ -45,11 +45,20 @@ pub(crate) struct InterierSources {
...
             sources: Vec::new(),
-        }
+        };
+
+        // Here be dragons!
+        // In order to reduce the span size, we allocate
+        // a dummy source at the beginning. This means
+        // we don't need to use an Option<u32>, but rather
+        // just a plain old u32 for the source id.
+        sources.add_source("", None);
```
mattwparas, line 59: This probably can just go back to being an `Option<u32>` after cleaning this up some more. I thought I was going to end up saving space but after working on it more I don't think that is the case.

crates/steel-repl/src/repl.rs
```diff
@@ -219,6 +220,9 @@ pub fn repl_base(mut vm: Engine) -> std::io::Result<()> {
                     ":pwd" => println!("{current_dir:#?}"),
                     // ":env" => vm.print_bindings(),
                     ":?" | ":help" => display_help(),
+                    ":interner" => {
+                        println!("{}", interned_current_memory_usage())
+                    }
```
mattwparas, line 225: TODO: Remove this

PR #252, opened 2024-07-20
update steel/ord

fominok:
Hey,
I'm not sure what is the goal regarding consistency with Scheme, but according to Racket docs for `>`, `<`, `>=` and `<=` functions both args shall be `real?`. Also things like `(< 1337 "leet")` didn't fail either.

This PR updates `steel/ord` module for real numbers comparison as well as adds missing documentation.

P.S. I tried to implement something like numeric tower as well to promote types in some more typesafe way to possibly reuse it in other modules, but failed miserably, shall be a separate effort I suppose, going the most direct way now.

### DIFF
--- a/crates/steel-core/Cargo.toml
+++ b/crates/steel-core/Cargo.toml
@@ -68,6 +68,8 @@ anyhow = { version = "1", optional = true }
 
 stacker = { version = "0.1.15", optional = true }
 
+bigdecimal = "0.4.5"
+
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 getrandom = { version = "*", features = ["js"] }
 
--- a/crates/steel-core/src/steel_vm/primitives.rs
+++ b/crates/steel-core/src/steel_vm/primitives.rs
@@ -61,9 +61,7 @@ use crate::{
     rvals::{Result, SteelVal},
     SteelErr,
 };
-use bigdecimal::BigDecimal;
 use fxhash::{FxHashMap, FxHashSet};
-use num::{bigint::ToBigInt, BigRational, FromPrimitive};
 use once_cell::sync::Lazy;
 use std::{cell::RefCell, cmp::Ordering};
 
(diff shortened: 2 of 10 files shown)


fominok:
Unfortunately I can't run all tests locally to see what's going on, but my not so wild guess is that ord tests exist for strings and for numbers and both use same functions, so fix for strings to use char comparison functions breaks the same test that is fed with numbers; i don't know what the final decision here, shall `>` be generic as possible or we still take inspiration from Racket?

mattwparas:
To answer the notion that `<` should only work with numbers, I'm inclined to agree with that sentiment, matching racket and guile would be a good idea then

mattwparas:
Also, running the tests locally, seems like just a mixup - you did edit the `trie.rkt` but the test is running a different `trie` file - just update `cogs/sorting/trie-sort.scm` to with the `char<?` and the tests will pass

fominok:
> Also, running the tests locally, seems like just a mixup - you did edit the `trie.rkt` but the test is running a different `trie` file - just update `cogs/sorting/trie-sort.scm` to with the `char<?` and the tests will pass

it seems there are two trie-sort.scm in tests and I fixed only one besides trie.rkt, thanks

fominok:
Since it's passing, does it require some extra work to be merged?

mattwparas:
Sorry! Been a very busy few days. I will review it today

crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -949,13 +951,248 @@ fn equality_module() -> BuiltInModule {
     module
 }
 
+/// Real numbers ordering module.
+#[steel_derive::define_module(name = "steel/ord")]
 fn ord_module() -> BuiltInModule {
     let mut module = BuiltInModule::new("steel/ord");
+
+    fn steel_ord(left: &SteelVal, right: &SteelVal) -> Result<Option<Ordering>> {
+        Ok(match (left, right) {
```
mattwparas, line 960: Not sure if you saw this, but this is how `PartialOrd` is implemented for steel values as of now: https://github.com/mattwparas/steel/blob/master/crates/steel-core/src/rvals.rs#L2081

Could we just lift the comparisons for numbers and re use them here?

fominok: Indeed I didn't see it, thanks

Did I understand correctly that you suggest to improve those existing ones?

mattwparas: Sorry for the delay, been a busy few days. I didn't do an exhaustive comparison, but mostly suggesting using the existing comparison code as reference to make sure we're not missing anything - whether that means improving what is there or for the stuff you've written. I'll do another pass shortly to see what the differences are if any

fominok: Hey, i've reused existing PartialEq since it's there, and updated it as well in order not to loose precision easily, per TODO above

crates/steel-core/Cargo.toml
```diff
@@ -68,6 +68,8 @@ anyhow = { version = "1", optional = true }
 
 stacker = { version = "0.1.15", optional = true }
 
+bigdecimal = "0.4.5"
```
mattwparas, line 71: Is this additional dependency needed?

fominok: I used bigdecimals to cover cases with float and something possibly bigger than f64, like BigRational

mattwparas: I'm not explicitly against bringing in another dependency, but what was the behavior before? Were we just losing precision then? I'm not sure if scheme even has big floats as a data type in general so I'm not explicitly sure if this is necessary

fominok: Yes, precision, and need to manually handle Big* types conversion if they can't fit in f64 if we do without BigDecimal, to shortcut to "greater than"

mattwparas: Got it, I think this is fine then. Sorry for the delay once again

crates/steel-core/src/steel_vm/primitives.rs
```diff
@@ -949,13 +952,116 @@ fn equality_module() -> BuiltInModule {
...
+    ///
+    /// # Examples
+    /// ```scheme
+    /// > (> 3 2) ;; => #t
+    /// > (> 1 1) ;; => #f
+    /// > (> 3/2 1.5) ;; => #f
+    /// > (> 3/2 1.4) ;; => #t
+    /// ```
+    #[steel_derive::function(name = ">")]
+    fn greater_than(left: &SteelVal, right: &SteelVal) -> Result<SteelVal> {
```
mattwparas, line 975: So the only other comment I'd have is that these functions are typically multi-arity - as in they take an arbitrary number of parameters. Could we use the rest args here to implement that? That would apply to all of the primitives here as well

crates/steel-core/Cargo.toml
fominok: No worries, thanks!

crates/steel-core/src/steel_vm/primitives.rs
fominok: Hey, I've updated the PR; One thing I noticed though is that `arty = "AtLeast(1)"` seems to do nothing and proceeds with things like `(>)`

mattwparas: I believe the arity annotation only comes into play with documentation, so it won't enforce any checks

mattwparas: So in this case - the only thing remaining is to add an arity check that we're passing at least one argument in (which I think is the behavior we want) - otherwise this LGTM

fominok: Got it, updated the check

Then I think we're good to go

mattwparas (approved):

PR #261, opened 2024-08-19
simplify `flake.nix`

MatrixManAtYrService:
I'll confess to not really knowing what's going on here.  I just saw that there was a flake.nix here, so I dropped into a devshell and got an error about a hash mismatch. This PR updates it to the "correct" hash according to my experience.

Thought you might like to know, sorry I can't provide more context--I'm new to this project, although I'm looking forward to getting to know it better.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,10 +34,6 @@
           };
           cargoLock = {
             lockFile = ./Cargo.lock;
-            # Temporary fix until https://github.com/mattwparas/steel/issues/192 is fixed upstream
-            outputHashes = {
-              "lasso-0.7.2" = "sha256-ccqcDvWZD5hp4iZ420jgNJtR+MaVlqHFtXU2GWkbyfg=";
-            };
           };
           cargoBuildFlags = "-p cargo-steel-lib -p steel-interpreter";
           buildInputs = [openssl] ++ lib.optionals stdenv.isDarwin [darwin.apple_sdk.frameworks.Security];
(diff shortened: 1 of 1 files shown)


mattwparas:
thanks! I don't use nix so these issues seem to pop up every once in a while. Perhaps I should start using it

MatrixManAtYrService:
Nix has its perks, and its pitfalls.  I think it's worth it :man_shrugging: 

It looks like the hash has some context here: https://github.com/mattwparas/steel/pull/203#discussion_r1596263667, I don't know cargo or lasso well enough to weigh in on the specifics.

mattwparas:
So it looks like the lasso dependency now has an up to date cargo dependency, which would mean we wouldn't need this has anymore.

MatrixManAtYrService:
That's good news 🙂 

I'm not a skilled cargo user, so double-check me here, but I've changed the two `Cargo.toml` files which reference lasso like so:

```diff
- lasso = { git = "https://github.com/kixiron/lasso/", features = ["multi-threaded", "serialize"] }
+ lasso = { version = "0.7.3", features = ["multi-threaded", "serialize"] }
```

Then I deleted `Cargo.lock` and ran `cargo generate-lockfile`.  Following that, the following commands all seemed to work ok:

- `nix build` (used to break with a hash mismatch)
- `cargo build`
- `cargo test`

Are there any additional checks that are warranted here?

MatrixManAtYrService:
That's good news. After rebasing I find that I can remove the hash and `nix build` still completes without trouble 😄

mattwparas (approved):

PR #262, opened 2024-08-20
implement vector literals

jrvidal:
Implements vector and bytevector literals.

## TODO
* No vector patterns for macros. I'll probably work on it on a future PR.

### DIFF
--- a/crates/steel-core/src/compiler/code_gen.rs
+++ b/crates/steel-core/src/compiler/code_gen.rs
@@ -907,6 +907,17 @@ impl<'a> VisitorMut for CodeGenerator<'a> {
 
         Ok(())
     }
+
+    fn visit_vector(&mut self, v: &crate::parser::ast::Vector) -> Self::Output {
+        for arg in &v.args {
+            self.visit(arg)?;
+        }
+
+        let payload_size = 2 * v.args.len() + usize::from(v.bytes);
+        self.push(LabeledInstruction::builder(OpCode::VEC).payload(payload_size));
+
+        Ok(())
+    }
 }
 
 #[cfg(test)]
--- a/crates/steel-core/src/compiler/code_gen.rs
+++ b/crates/steel-core/src/compiler/code_gen.rs
@@ -20,13 +20,17 @@ use crate::{
         parser::SyntaxObject,
         span_visitor::get_span,
         tokens::TokenType,
+        tryfrom_visitor::TryFromExprKindForSteelVal,
         visitors::VisitorMut,
     },
     stop, SteelVal,
 };
 use smallvec::SmallVec;
 use std::sync::atomic::AtomicUsize;
-use steel_parser::tokens::{IntLiteral, NumberLiteral, RealLiteral};
+use steel_parser::{
+    ast::Quote,
+    tokens::{IntLiteral, NumberLiteral, RealLiteral},
+};
 
 use crate::rvals::Result;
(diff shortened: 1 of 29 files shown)


crates/steel-parser/src/ast.rs
```diff
@@ -977,6 +997,60 @@ impl From<Require> for ExprKind {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct Vector {
+    pub args: Vec<ExprKind>,
+    pub bytes: bool,
+    pub span: Span,
+}
```
jrvidal, line 1005: A single struct for both vectors and bytevectors. It's not too cumbersome and saves us from adding a `visit_bytevector()` method in a zillion places.

crates/steel-core/src/compiler/code_gen.rs
```diff
@@ -907,6 +907,17 @@ impl<'a> VisitorMut for CodeGenerator<'a> {
 
         Ok(())
     }
+
+    fn visit_vector(&mut self, v: &crate::parser::ast::Vector) -> Self::Output {
+        for arg in &v.args {
+            self.visit(arg)?;
+        }
+
+        let payload_size = 2 * v.args.len() + usize::from(v.bytes);
```
jrvidal, line 916: The payload for `VEC` is a mixture of length and byte-ness, with a bit of bit-fiddling.

mattwparas (commented):
This is awesome, thank you! 

The code LGTM - could we get some tests on the vector literals? I think there are some for bytevector literals, but I didn't see any on non byte vector literals (correct me if I'm wrong please)

jrvidal:
Funny how little I know about Scheme :upside_down_face: My initial implementation was totally wrong, b/c vectors were not self-quoting, so `#(a)` would throw if the `a` variable were undefined, instead of giving me the `a` symbol.

Vectors are valid quasiquote patterns, but this is not working yet, but I suppose I need to add vector macro patterns first. I also think a bunch of `visit_vector()` implementations are redundant or even incorrect, once you realize that the contents of the vector are "inert". I'll revisit that in the macro PR.

crates/steel-core/src/primitives/bytevectors.rs
```diff
@@ -289,23 +289,24 @@ pub fn list_to_bytes(value: Vec<u8>) -> Result<SteelVal> {
     Ok(SteelVal::ByteVector(SteelByteVector::new(value)))
 }
 
-/// Append two byte vectors into a new bytevector.
+/// Append multiple byte vectors into a new bytevector.
 ///
 /// # Examples
 /// ```scheme
-/// (bytes-append (bytes 0 1 2) (bytes 3 4 5)) ;; => (bytes 0 1 2 3 4 5)
+/// (bytes-append #u8(0 1 2) #u8(3 4 5)) ;; => #u8(#x00 #x01 #x02 #x03 #x04 #x05)
```
jrvidal, line 296: I suppose all docs could be updated to use literals like this one. I might get around doing it in this PR or a subsequent one.

jrvidal:
Oops, I'll fix those tests.

mattwparas:
Once the sync work lands that whole type moving across threads stuff will go away, so that's nice at least

mattwparas (approved):

mattwparas:
Thank you!

PR #266, opened 2024-09-04
strip the release binary

kskarthik:
set option to strip the binary in release mode. Helps reduce the binary size as well as the docker image size


kskarthik:
looks like strip is the default option & debug false overrides it

PR #271, opened 2024-10-09
Added change-current-directory and parent-name to fs

Talia-12:
This is a simple PR adding a couple methods for fs interactions that I was missing, namely a function to change the current working directory and a function mirroring file-name that returns the full path to the parent directory. I also attempted to add the correct docs for them, but it's possible I missed some places (and if there's automatic doc generation I didn't know how to run it).


mattwparas:
This LGTM - but it looks like rust fmt is failing. Would you be able to re format it and push? Otherwise I can handle it

Talia-12:
Okay, I think I fixed it?

mattwparas (approved):

PR #279, opened 2024-10-26
Clean up flake.nix

bloxx12:
This PR gets rid of the flake's dependency on `flake-utils`, which is an  abstraction library which is definitely not necessary here. This might be nice to reduce the flake's complexity and make it more accessible for non-`flake-utils` users. 

Furthermore, I've moved the `devShell` and `steel package` parts of the flake into seperate files, which has several benefits:
- Upstreaming steel into `Nixpkgs` becomes much easier
- Common in many projects which have a `flake.nix` 
- Reduces the flake's complexity by a bit

I still have to test this to make sure everything behaves the same way as before, but it should be more or less correct.

This is a draft, so I'm open to any changes ofc  :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -13,7 +13,10 @@
   in {
     packages = eachSystem (system: {
       default = self.packages.${system}.steel;
-      steel = pkgsFor.${system}.callPackage ./nix/package.nix {};
+      steel = pkgsFor.${system}.callPackage ./nix/package.nix {
+        inherit (self.packages.${system}) steel;
+        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) CoreServices SystemConfiguration;
+      };
     });
 
     formatter = eachSystem (system: pkgsFor.${system}.alejandra);
@@ -21,7 +24,9 @@
     defaultPackage = eachSystem (system: self.packages.${system}.default);
 
     devShells = eachSystem (system: {
-      default = pkgsFor.${system}.callPackage ./nix/package.nix {};
+      default = pkgsFor.${system}.callPackage ./nix/package.nix {
+        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) Security;
+      };
     });
 
     apps = eachSystem (system: {
--- a/flake.nix
+++ b/flake.nix
@@ -1,14 +1,17 @@
 {
   description = "Steel";
 
-  inputs.nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    systems.url = "github:nix-systems/default";
+  };
 
   outputs = {
     self,
     nixpkgs,
+    systems,
   }: let
-    systems = ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"];
-    eachSystem = nixpkgs.lib.genAttrs systems;
+    eachSystem = nixpkgs.lib.genAttrs (import systems);
     pkgsFor = nixpkgs.legacyPackages;
   in {
     packages = eachSystem (system: {
@@ -19,6 +22,8 @@
     });
 
     formatter = eachSystem (system: pkgsFor.${system}.alejandra);
+
+    # DEPRECATED
     legacyPackages = self.packages;
     defaultPackage = eachSystem (system: self.packages.${system}.default);
 
@@ -32,7 +37,9 @@
     apps = eachSystem (system: {
       steel = {
         type = "app";
-        program = "${self.packages.${system}.steel}/bin/steel";
+        program =
+          pkgsFor.${system}.lib.getExe
+          self.packages.${system}.steel;
       };
       default = self.apps.${system}.steel;
     });
--- a/flake.nix
+++ b/flake.nix
@@ -14,8 +14,7 @@
     packages = eachSystem (system: {
       default = self.packages.${system}.steel;
       steel = pkgsFor.${system}.callPackage ./nix/package.nix {
-        inherit (self.packages.${system}) steel;
-        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) CoreServices SystemConfiguration;
+        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) Security;
       };
     });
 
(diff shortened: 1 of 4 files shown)


mattwparas:
Thanks for this! Unfortunately I don't use nix - any recommendations on how to test this?

bloxx12:
Hey, sorry for the late reply. I just tested everything and it should all just work. :)

mattwparas:
> Hey, sorry for the late reply. I just tested everything and everything works properly :)

Great, no problem. What should I run to test this on my end? Also, is there a way I can test in the CI?

bloxx12:
The original flake.nix exposes several things, namely:
1. A formatter
`      formatter = pkgs.alejandra;
`
2. The steel package
`      packages.steel = steel;
`
3. a default package (steel)
`      defaultPackage = packages.steel;
`
4. a devShell
`      devShell = with pkgs;
...
`
 5. a steel app
 `      apps.steel = {
        type = "app";
        program = "${steel}/bin/steel";
      };
`
 6. a default app (steel)
 `      apps.default = apps.steel;
`
The flake-utils library provides an abstraction over different architectures to make it a bit easier to provide all this for all architectures, although this can be just done like I've done here, using eachSystem:
alejandra is one of the standard nix formatters.
`    formatter = eachSystem (system: pkgsFor.${system}.alejandra);
`
which evaluates to:
```nix
formatter = {
	x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.alejandra
	aarch64-linux = nixpkgs.legacyPackages.aarch64-linux.alejandra
	x86_64-darwin = nixpkgs.legacyPackages.x86_64-darwin.alejandra
	aarch64-darwin = nixpkgs.legacyPackages.aarch64-darwin.alejandra
```
	
formatter

I've done this for basically everything this flake exposes.
`legacyPackages = packages;` is basically redundant here but I've kept it to not change too much.

Now, to test these changes, there are several things which need to be tested:
(in the steel directory:)
- `nix build .` (builds the default package)
- `nix build .#steel` (builds the steel package)
- `nix run .` (builds and runs the default package)
- `nix run .#steel` (builds and runs the steel package)
- `nix develop .` (creates a devShell) 

I've sucessfully ran all of them and you should be able to too if you have nix installed.

As for the CI, I'm positive that's possible. I'll research a bit and comment later again. 
If you have any more questions I'd be glad to answer them :)

Dragyx (approved):
Seems to do exactly what it did before, just without flake-utils.
Separating the package and dev shell code into separate files is nice as well.

flake.nix
```diff
@@ -1,87 +1,40 @@
...
   outputs = {
     self,
     nixpkgs,
-    flake-utils,
-  }:
-    flake-utils.lib.eachDefaultSystem (system: let
-      pkgs = nixpkgs.legacyPackages.${system};
-      fs = pkgs.lib.fileset;
+  }: let
+    systems = ["x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin"];
```
fufexan, line 10: Add `inputs.systems.url = "github:nix-systems/default";` and use `systems = import systems;` instead.

This way users can override systems to try and build on other platforms.

flake.nix
```diff
@@ -1,87 +1,40 @@
...
+    packages = eachSystem (system: {
+      default = self.packages.${system}.steel;
+      steel = pkgsFor.${system}.callPackage ./nix/package.nix {
+        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) Security;
+      };
+    });
+
+    formatter = eachSystem (system: pkgsFor.${system}.alejandra);
+    legacyPackages = self.packages;
+    defaultPackage = eachSystem (system: self.packages.${system}.default);
```
fufexan, line 23: Add a comment above these two that says `# DEPRECATED`. `legacyPackages` doesn't seem to need to be here, as the package set is flat, so it's already supported by `packages`.

`defaultPackage` is superseded by `packages.default`, which you already added.

flake.nix
```diff
@@ -1,87 +1,40 @@
...
+        inherit (self.packages.${system}) steel;
+        inherit (pkgsFor.${system}.darwin.apple_sdk.frameworks) CoreServices SystemConfiguration;
+      };
+    });
+
+    apps = eachSystem (system: {
+      steel = {
         type = "app";
-        program = "${steel}/bin/steel";
+        program = "${self.packages.${system}.steel}/bin/steel";
```
fufexan, line 35:
```suggestion
        program = lib.getExe self.packages.${system}.steel;
```

nix/package.nix
fufexan, line 1: Newlines between each major binding in this file could go a long way in improving readability.
e.g. `pname` and `version` should be separated from `src` by a newline, same with `*inputs`.

nix/package.nix
```diff
@@ -0,0 +1,55 @@
...
+      substituteInPlace cogs/installer/download.scm --replace-warn "cargo-steel-lib" "$out/bin/cargo-steel-lib"
+      mkdir $out/lib
+      export STEEL_HOME="$out/lib"
+      pushd cogs
+      $out/bin/steel install.scm
+      popd
+      rm "$out/bin/cargo-steel-lib"
+    '';
+    meta = {
+      description = "An embedded scheme interpreter in Rust";
```
fufexan, line 50:
```suggestion
      description = "Embedded scheme interpreter in Rust";
```

nix/package.nix
```diff
@@ -0,0 +1,55 @@
...
+      homepage = "https://github.com/mattwparas/steel";
+      license = with lib.licenses; [asl20 mit];
+      mainProgram = "steel";
```
fufexan, line 53: Could also add `platforms = lib.platforms.unix;`

nix/shell.nix
```diff
@@ -0,0 +1,31 @@
...
+    [cargo rustc openssl libiconv]
+    ++ lib.optionals stdenv.isDarwin [
+      CoreServices
+      SystemConfiguration
+    ];
+  nativeBuildInputs = [
+    pkg-config
+    rust-analyzer
+    rustfmt
+  ];
```
fufexan, line 30:
```suggestion
  packages = [
    cargo
    rustc
    rust-analyzer
    rustfmt
    libiconv
  ] ++ lib.optionals stdenv.isDarwin [
    CoreServices
    SystemConfiguration
  ];
  
  inputsFrom = steel;
```

`packages` should be used in shells, as the logical difference between `nativeBuildInputs` and `buildInputs` no longer applies.

`inputsFrom` takes pre-existing `nativeBuildInputs` and `buildInputs` from the given package and applies them to the shell.

fufexan (approved):

mattwparas (approved):

mattwparas:
Thank you all for this; I don't daily drive nix so it's helpful to have others testing this regularly and improving. Perhaps at some point I'll start using nix 😄

PR #286, opened 2024-11-05
analysis: don't do analysis for empty list

merisbahti:
This caused panic for the input `()`.

I was using `hx` with `steel-language-server`

```
2024-11-05T14:58:48.185 helix_lsp::transport [ERROR] steel-language-server err <- "thread 'main' panicked at crates/steel-core/src/compiler/passes/analysis.rs:1099:28:\n"
```


mattwparas:
Thank you!

PR #302, opened 2025-01-06
chore: fix low hanging lints

gibbz00:
This PR is the first (of probably many) that aims to bring steel closer to being able to run clippy as part of the CI.
Each crate will probably need its own PR for their remaining lint warnings and suggestions.

- Removes all those with level "error".
- Removes `#[allow(unused)]` annotations
- Runs a workspace-wide `cargo fix`

### DIFF
--- a/crates/cargo-steel-lib/src/lib.rs
+++ b/crates/cargo-steel-lib/src/lib.rs
@@ -27,10 +27,8 @@ pub fn steel_home() -> Option<PathBuf> {
                 // this is probably not the best place to do this. This almost
                 // assuredly could be lifted out of this check since failing here
                 // could cause some annoyance.
-                if !x.exists() {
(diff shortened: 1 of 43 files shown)


mattwparas:
Just checking - the intent is here is to keep the unused warnings? Is that happening as a part of the next PR?

gibbz00:
The intent was to remove most of the unused code, only to then realize that it was a lot (600+ clippy warnings, excluding commented out code).

I can remove that commit for now if it makes more sense.

mattwparas:
Got it - I didn't realize I was that far behind on clippy warnings :sweat_smile: - so this does need to get addressed eventually.

The next MR starts to address the unused code right? So in theory the warnings living on master wouldn't live for too long?

Also, could we leave the unused in `vm.rs` - that should help a bit. I'll clean that one up myself since its a bit of a mess

PR #307, opened 2025-02-10
rework unicode escapes

jrvidal:
# What other implementations support

### Rust

Valid in both strings and characters

* `\xDD` (only up to `\x7F`)
* `\u{D}` -  `\u{DDDDDD}`
  (enforces valid codepoints, no unpaired surrogates. e.g. `'\u{d83d}'` does not compile)

### Racket

* `#\uD` - `#\uDDDD`
* `#\UD` - `#\UDDDDDD`
* `"\xD"`, `"\xDD"`
* `"\uD"` - `"\uDDDD"`
* `"\UD"` - `"\UDDDDDD"`

They all enforce valid codepoints. You can also write paired surrogates directly: "\uDDDD\uDDDD" (where each individual codepoint is not valid on its own).

### R7RS

* `#\xD...`
* `"\xD...;"`

Character set is not specc'ed, so there's no validation _a priori_. But for a Unicode-based implementation, it's kinda obvious.

# What this PR has

1. ~`"\xDD"`~ (removed)
1. `"\xD...;"` (new)
1. `"\uD...;"` (new)
1. `"\u{D...}"` (new)
1. `#\xD...` (new)
1. `#\uD...` (new-ish, any number of digits accepted now)
1. `#\u{D...}` (new-ish, any number of digits accepted now)

(Only valid Unincode codepoints are accepted)

The main point of this PR is to remove (1) and to add (2) and (5). I added (3) and (4) only for consistency, but I'm far from married to them. I think it makese sense that, if we keep (6) and (7), they should not have length restrictions.

### DIFF
--- a/crates/steel-parser/src/lexer.rs
+++ b/crates/steel-parser/src/lexer.rs
@@ -952,6 +952,33 @@ mod lexer_tests {
     }
 
     #[test]
+    fn test_string_newlines() {
+        let mut s = TokenStream::new(
+            " \"foo\nbar\" \"foo \\  \n   bar\" ",
+            true,
+            SourceId::none(),
+        );
+
+        assert_eq!(
+            s.next().unwrap(),
+            Token {
+                ty: StringLiteral(Arc::from("foo\nbar".to_string())),
+                source: "\"foo\nbar\"",
+                span: Span::new(1, 10, SourceId::none())
+            }
+        );
+
+        assert_eq!(
+            s.next().unwrap(),
+            Token {
+                ty: StringLiteral(Arc::from("foo bar".to_string())),
+                source: "\"foo \\  \n   bar\"",
+                span: Span::new(11, 27, SourceId::none())
+            }
+        );
+    }
+
+    #[test]
     fn test_unexpected_char() {
         let mut s = TokenStream::new("($)", true, SourceId::none());
         assert_eq!(
--- a/crates/steel-parser/src/lexer.rs
+++ b/crates/steel-parser/src/lexer.rs
@@ -1,7 +1,8 @@
 use super::parser::SourceId;
-use crate::tokens::{parse_unicode_str, NumberLiteral, Paren, ParenMod, RealLiteral};
 use crate::tokens::{IntLiteral, Token, TokenType};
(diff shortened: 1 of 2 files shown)


crates/steel-parser/src/lexer.rs
```diff
@@ -110,74 +127,120 @@ impl<'a> Lexer<'a> {
...
-
-                        let char = char::from_u32(byte as u32)
-                            .ok_or_else(|| TokenError::MalformedByteEscape)?;
+                        let char =
+                            char::from_u32(codepoint).ok_or(TokenError::MalformedByteEscape)?;
 
                         buf.push(char);
                     }
 
+                    Some(&start @ (' ' | '\t' | '\n')) => {
```
jrvidal, line 169: Also added non-significant whitespace for strings:

```
"foobar   \
    baz"
=> "foobarbaz"
```

crates/steel-parser/src/lexer.rs
```diff
@@ -802,6 +869,88 @@ mod lexer_tests {
...
+            r#" "\u0045" "#,
+            r#" #\xaaaaaaaa " "#,
+            r#" "\u{ffffffff}" "#,
+            r#" #\u{} "#,
+        ];
+
+        for token in tokens {
+            let mut s = TokenStream::new(token, true, SourceId::none());
+
+            assert_eq!(s.next().unwrap().ty, Error, "{:?} should be invalid", token);
```
jrvidal, line 950: I'd like to revisit tokenization errors at some point. They surface very little info in error messages.

mattwparas:
Thanks for this - looks like there is a flaky test in the build. I'll look into that

crates/steel-parser/src/lexer.rs
mattwparas: So this removes the whitespace within the string following the `foobar`? Wouldn't we want that to be included?

jrvidal: Which makes me realize that this needs a test.

mattwparas (approved):

jrvidal:
Shall we merge this? I'm not sure if we wanted to reconsider anything.

mattwparas:
Looks like I approved but forgot to hit the merge button

PR #310, opened 2025-02-18
support escaped identifiers

jrvidal:
adds support for escaped identifiers such as `|foo bar|`.

### DIFF
--- a/crates/steel-parser/src/lexer.rs
+++ b/crates/steel-parser/src/lexer.rs
@@ -406,9 +406,9 @@ impl<'a> Lexer<'a> {
                 '\\' if escaped_identifier => {
                     self.eat();
 
-                    if let Some(c) = self.read_string_escape()? {
-                        ident_buffer.push_escape(c);
-                    }
+                    let escaped = self.read_string_escape()?;
+
+                    ident_buffer.push_escape(escaped);
                 }
                 c if escaped_identifier => {
                     ident_buffer.push(c);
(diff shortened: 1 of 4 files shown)


crates/steel-parser/src/lexer.rs
```diff
@@ -16,13 +17,13 @@ pub const NEG_NAN: &str = "-nan.0";
 pub struct OwnedString;
 
 impl ToOwnedString<String> for OwnedString {
-    fn own(&self, s: &str) -> String {
+    fn own(&self, s: Cow<str>) -> String {
```
jrvidal, line 20: This `Cow` epidemic is a bit noisy.

crates/steel-parser/src/lexer.rs
```diff
@@ -327,29 +343,29 @@ impl<'a> Lexer<'a> {
...
+            _ => self.read_word(),
         }
     }
 
-    fn read_number(&mut self) -> TokenType<&'a str> {
+    fn read_number(&mut self) -> Result<TokenType<Cow<'a, str>>> {
         while let Some(&c) = self.chars.peek() {
             match c {
-                c if c.is_numeric() => {
+                c if c.is_ascii_digit() => {
```
jrvidal, line 353: :warning: Unrelated fix.

crates/steel-parser/src/lexer.rs
```diff
@@ -455,6 +509,42 @@ impl<'a> Lexer<'a> {
     }
 }
 
+struct IdentBuffer<'a> {
```
jrvidal, line 512: This is a small helper to optimize all those `Cow`s. We only allocate a buffer for the identifier if we encounter an explicit escape.

mattwparas (approved):
LGTM - sorry for the delay

PR #312, opened 2025-02-23
Repl: Colored prompt on windows

SpecificProtagonist:
`colored` supports Windows (by calling `SetConsoleMode`) and is already used to display the greeting.


mattwparas:
Thanks - I originally had removed the colored prompt because of an odd issue with how the prompt rendered on windows.
Were you able to confirm that the prompt works as expected on windows?

PR #313, opened 2025-02-24
Repl: completions for globals

SpecificProtagonist:
I'm trying to use Steel to learn Scheme, but I don't know what functions are available, so this PR makes the repl tell me.

As I don't yet know all the naming conventions in Scheme & Steel, there may be a better priorisation order for the completions.

I've kept this in the `highlight` module as it defines the `RustylineHelper`, but it might be worth pulling it out into a separate small module.

This impl doesn't complete names of macros; I'm not sure yet how they're handled – they aren't highlighted currently.


mattwparas:
This is great, thank you!

mattwparas (approved):

PR #315, opened 2025-02-24
Repl: Include parse errors in validation

SpecificProtagonist:
Catching parse errors in validation is less disruptive than clearing the line. Ideally all compile errors should be validated, but that currently requires a mutable ref to the engine, which I don't think we can get during a safepoint?

### DIFF
--- a/crates/steel-repl/src/highlight.rs
+++ b/crates/steel-repl/src/highlight.rs
@@ -1,7 +1,8 @@
 extern crate rustyline;
 use colored::Colorize;
+use steel::SteelErr;
 use steel_parser::interner::InternedString;
-use steel_parser::parser::SourceId;
+use steel_parser::parser::{Parser, SourceId};
 use steel_parser::tokens::TokenType;
 
 use std::collections::HashSet;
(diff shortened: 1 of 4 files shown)


mattwparas:
Correct - we can't get a mutable ref to the engine during the safepoint, let me test out this branch to see the behavior.

crates/steel-repl/src/highlight.rs
```diff
@@ -67,6 +68,11 @@ impl Validator for RustylineHelper {
 
         if balance > 0 || unfinished {
             Ok(ValidationResult::Incomplete)
+        } else if let Some(parse_err) = Parser::new(ctx.input(), None).find_map(|r| r.err()) {
+            Ok(ValidationResult::Invalid(Some(format!(
+                "\n{}",
+                SteelErr::from(parse_err).emit_result_to_string("", "", true)
+            ))))
```
mattwparas, line 75: Sorry for getting to this late - I like this change and would like to include it! However, could you check what happens when you type:

```
(define ##foo 10)
```

And then hit enter? For me, the repl won't accept it - I think this check also has to be done after line 61 since now the double hash identifier is a parser error.

Otherwise, this change LGTM if you're still interested in getting it in.

mattwparas:
One small comment - apologies for the delay on it

PR #316, opened 2025-02-26
Improve nix package and shell

HeitorAugustoLN:
Also added support for nix without flakes


mattwparas:
Forgive me for the delay on reviewing this - I don't really know nix 😅 - let me do a little research before I make any comments

tesujimath:
I had a look at this.  It suffers from the same problem as #353 which I fixed in #356.

Please could you incorporate that fix?  Then I'll have a further look at this. :smile:

tesujimath:
Does this break the build on darwin?  I see the darwin specific Security was removed.

Perhaps we need a Nix workflow in GitHub actions?  I'll make an issue!

HeitorAugustoLN:
> Does this break the build on darwin? I see the darwin specific Security was removed.

No, it does not, I've tested it on darwin

HeitorAugustoLN:
> Please could you incorporate that fix? Then I'll have a further look at this. 😄

Done!

tesujimath:
In order to get this to work when I [rebased it onto upstream master](https://github.com/tesujimath/steel/commits/improve-package-rebased/), I had to switch from `cargoHash` to `cargoLock`, and remove `useFetchCargoVendor`

See the discussion in the [Nix docs for Rust packaging](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/rust.section.md#importing-a-cargolock-file-importing-a-cargolock-file).

`nix/package.nix`:
```
@@ -27,8 +24,9 @@ rustPlatform.buildRustPackage {
     fileset = lib.fileset.gitTracked ../.;
   };
 
-  useFetchCargoVendor = true;
-  cargoHash = "sha256-8S/CJFXv89lS3Gyd/TgHUrMxmvYQZvbXCiPQcQtL3Jo=";
+  cargoLock = {
+    lockFile = ../Cargo.lock;
+  };
```

Also, not breaking, but parameters `stdenv` and `darwin` are unused.

With this change, I am able to depend on Steel via a Nix flake in my playpen repo, which is super-awesome.  Thanks!

tesujimath (approved):
Thanks, this works for me, I'm happy with this now. :smile:

mattwparas (approved):

mattwparas:
Thanks for team work on reviewing this!

PR #318, opened 2025-03-02
Added documentation to numbers.rs

gold-silver-copper:
I added documentation to the remainder operator, and the trigonometry operators. Modified the modulo operator doc a bit. Added a TODO to exact operator.

This is my first contribution to this project, please let me know if there's anything I can do better. Thank you, I look forward to contributing more in the future.

### DIFF
--- a/crates/steel-core/src/primitives/numbers.rs
+++ b/crates/steel-core/src/primitives/numbers.rs
@@ -383,6 +383,7 @@ pub fn quotient(args: &[SteelVal]) -> Result<SteelVal> {
 }
 
 /// Returns the euclidean remainder of the division of the first number by the second
+/// This differs from the remainder operator when using negative numbers.
 ///
 /// (modulo n m) -> integer?
 ///
@@ -433,7 +434,19 @@ pub fn remainder(args: &[SteelVal]) -> Result<SteelVal> {
     }
 }
 
-// TODO: Do this for sin, cos, tan, asin, acos, atan
+/// Returns the sine value of the input angle, measured in radians.
+///
+/// (sin n) -> number?
+///
+/// * n : number? - The input angle, in radians.
+///
+/// # Examples
+/// ```scheme
+/// > (sin 0) ;; => 0
+/// > (sin 1) ;; => 0.8414709848078965
+/// > (sin 2.0) ;; => 0.9092974268256817
+/// > (sin 3.14) ;; => 0.0015926529164868282
+/// ```
 #[steel_derive::function(name = "sin", constant = true)]
 pub fn sin(arg: &SteelVal) -> Result<SteelVal> {
     match arg {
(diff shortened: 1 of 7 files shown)


gold-silver-copper:
I will work on adding docs to other files in separate PRs for now.

crates/steel-core/src/primitives/numbers.rs
```diff
@@ -562,6 +655,7 @@ pub fn exactp(value: &SteelVal) -> bool {
     }
 }
 
+/// TODO ?
```
mattwparas, line 658: I'm fine to leave the todo in - but can you make it `//` instead of a doc comment `///` - otherwise it'll show up as todo in the docs itself

mattwparas:
When you're happy with the docs, can you also run `cargo xtask docgen` to generate the docs that go on the site? I have those get checked in manually for now. I'd like to get them to be generated by the CI, but I haven't gotten around to that quite yet.

gold-silver-copper:
I added docs to the exact operator, and generated docs for the site. All the exposed functions in numbers.rs are now documented... until new ones get added. I think this PR is ready.

mattwparas:
Yeah, if they don't have the macro the docs won't get picked up, so some of the older function definitions haven't been converted to use that yet

mattwparas:
> I added docs to the exact operator, and generated docs for the site. All the exposed functions in numbers.rs are now documented... until new ones get added. I think this PR is ready.
> 
> Edit: I found a few undocumented functions in number.rs that didnt have a 'steel_derive::function(name=...)' decorator, such as even? and odd? . I tried adding docs above their function definitions and generating web docs, but it did not generate them properly.



> I added docs to the exact operator, and generated docs for the site. All the exposed functions in numbers.rs are now documented... until new ones get added. I think this PR is ready.
> 
> Edit: I found a few undocumented functions in number.rs that didnt have a 'steel_derive::function(name=...)' decorator, such as even? and odd? . I tried adding docs above their function definitions and generating web docs, but it did not generate them properly.

yeah, so those will need to be adjusted, the way they're bound to the module isn't the same as the way the other docs are added. No need to do this in this PR though

crates/steel-core/src/primitives/numbers.rs
```diff
@@ -562,6 +655,17 @@ pub fn exactp(value: &SteelVal) -> bool {
     }
 }
 
+/// Returns the input value if it is an exact number, otherwise raises an error.
```
mattwparas, line 658: Nit - I think this should say that this returns an exact representation of the input number, since it does coerce an inexact number to an exact form of it

mattwparas (changes requested):
One comment, otherwise LGTM

crates/steel-core/src/primitives/numbers.rs
gold-silver-copper: Thank you for the explanation, I changed the header and added a relevant example.

mattwparas (approved):

mattwparas:
Thank you!

PR #320, opened 2025-03-08
Moved functions from NumOperations to use derive | numbers.rs

gold-silver-copper:
I made all functions inside NumOperations use steel_derive macros instead. I then deleted the NumOperations struct and registered the functions. Also I moved the functions higher up in the file so that they are where the other function definitions are, above the preconditions.  Finally I added documentation to all the functions. 

Tests returned successfully. 

Notes: During my testing it seemed like Arity was not working. Setting Arity to variations of AtLeast and Exact did not affect how many inputs I could pass during run time.


mattwparas:
This LGTM - yep, the arity isn't checked. I'll take a look at your other PR that addresses that.

mattwparas (approved):

PR #321, opened 2025-03-08
Proof of Concept - Arity test code injection at compile time

gold-silver-copper:
When touching up the numbers.rs docs I noticed that the Arity values in the derive decorator do not actually make checks during run time. This was a bit confusing for me and so I thought it would be fun to try and add that myself.

What this PR does is determine the type of Arity (AtLeast,Exact, etc) and the integer value, then injects the proper if statement to do the arity check. 

Currently it only covers AtLeast and Exact, not Range or AtMost. I am rather inexperienced writing macros so it probably looks ugly. But overall of the concept, what do you think? Is this something the derive macro should do, or should Arity checks all be added manually. Let me know what you think.

Pros: Fixes bugs, maybe more intuitive? Currently running ```(log)``` on master will panic the repl, but on this branch it will give a proper error.

Cons: Code duplication, a lot of the functions already have arity checks, some of these could be removed, others which are inside match statements are harder to remove.


mattwparas:
Apologies for getting to this late - overall I like this change. I think moving the arity checks into the macro is probably a good idea since they should be happening anyway. Worst case, we can add a flag to the macro so skip adding the arity check or something if performance is a concern (which, I don't think it should be really).

gold-silver-copper:
Thanks for liking the change :D

I will work on further developing this then, things I may do are: 

1. Move the code for injection into a reusable function.
2. Remove existing arity checks from functions which will no longer need them.
3. Maybe try to cut down on code duplication inside the macros by moving parts into their own reusable functions.

gold-silver-copper:
This PR I think is ready, here are my notes for it:

1. I added a 'disable-arity-checking' feature to steel-derive and steel-core.
2. I imported 'builtin_stop!' into files which required it for the context macro arity checking.
3. Changed the arity value of some context functions, which were previously improperly set, such as in 'macro-case-bindings'
4. Enabled the 'extra-traits' feature for syn in the steel-derive crate. This provides needed types and traits for the arity checking injector.
5. Put duplicated code in steel-derive into a reusable function for 'native', 'native_mut', and 'context'.
6. Added a function that injects arity checking code into functions derived by 'native', 'native_mut', and 'context'.

All tests pass with the feature on or off.
PS: I started deduplicating code from 'function' and 'custom_function' but then realized they share exactly the same code...

crates/steel-derive/src/lib.rs
```diff
@@ -585,13 +586,100 @@ pub fn define_module(
...
+                    .expect("Arity value must be an integer"),
+            ))
+        })
+        .expect("Arity header is wrongly formatted");
+
+    //Context functions return values wrapped in Option, so we must use builtin_stop for them
+    let stop_type = if is_context_function {
+        quote! {builtin_stop!}
+    } else {
+        quote! {stop!}
```
mattwparas, line 638: Can we reference this like:

```rust
quote! { steel_core::builtin_stop! } 
```

I think if we do that then we don't have to import the macro at all the call sites (I could be wrong, but if it works that would be great)

gold-silver-copper: I will try this out and incorporate range into this PR soon.

gold-silver-copper: The following one did not work, due to a crate not being able to reference itself by name
```rust
quote! { steel_core::builtin_stop! } 
```
But this one did work:
```rust
quote! { crate::builtin_stop! } 
```

I think this is probably acceptable since it only affects Context functions.  What do you think?

gold-silver-copper:
Update: I implemented everything necessary to have Range arity checks. To test, I added the new Range header to the function 'range' (no pun intended). Seems to work on my end, let me know if you have other comments.

crates/steel-derive/src/lib.rs
```diff
@@ -585,13 +586,116 @@ pub fn define_module(
...
+    input: &ItemFn,
+    args: &Punctuated<Meta, Comma>,
+) -> (
+    proc_macro2::TokenStream,
+    proc_macro2::Ident,
+    String,
+    proc_macro2::Ident,
+    syn::Expr,
+    bool,
+) {
```
mattwparas, line 698: Only one comment - could we change this to return a struct which labels what each of those things are?

mattwparas (commented):
One comment, otherwise LGTM - thanks for this!

crates/steel-derive/src/lib.rs
gold-silver-copper: Done, I also renamed 'value' to 'steel_function_name' , and 'function_name' to 'rust_function_name', because I believe that to be more clear.

gold-silver-copper:
𓀡

mattwparas (approved):

PR #322, opened 2025-03-10
check for char equality in `eqv?` and `eq?`

m4rch3n1ng:
in section "6.1. Equivalence predicates" of [r7rs](https://standards.scheme.org/official/r7rs.pdf) it states:

The `eqv?` procedure returns `#t` if:
- [ ... ]
- *obj1* and *obj2* are both characters and are the same
character according to the `char=?` procedure.

slightly surprising may also be, that `eqv?` and `eq?` do the same thing, though it is not in violation of the scheme r7rs standard, which explicitly specifies, that `eq?` *may* return `#f` in some cases where `eqv?` returns `#t`.

this was previously not true about steel, but this fixes that.

before:
image

after:
image

### DIFF
--- a/crates/steel-core/src/rvals.rs
+++ b/crates/steel-core/src/rvals.rs
@@ -1754,6 +1754,7 @@ impl SteelVal {
             (IntV(l), IntV(r)) => l == r,
             (NumV(l), NumV(r)) => l == r,
             (BoolV(l), BoolV(r)) => l == r,
+            (CharV(l), CharV(r)) => l == r,
             (VectorV(l), VectorV(r)) => Gc::ptr_eq(&l.0, &r.0),
             (Void, Void) => true,
             (StringV(l), StringV(r)) => crate::gc::Shared::ptr_eq(l, r),


Cargo.toml
```diff
@@ -39,7 +39,7 @@ steel-derive = { path = "./crates/steel-derive", version = "0.5.0" }
 steel-repl = { path = "./crates/steel-repl", version = "0.6.0" }
 env_logger = "0.10.0"
 log = "0.4.17"
-clap = { version = "=4.1.4", features = ["derive"] }
+clap = { version = "4.1.4", features = ["derive"] }
```
mattwparas, line 42: what was this change for?

mattwparas:
Good catch! Thanks for this - I left one question, otherwise LGTM

Cargo.toml
m4rch3n1ng: my bad, that was by accident, tested something else and apparently accidentally commited parts of that here. reverted now

mattwparas (approved):

PR #323, opened 2025-03-11
implement `symbol=?` and fix symbol equality for `eqv?` and `eq?`

m4rch3n1ng:
implement the `symbol=?` procedure as specified in section "6.5 Symbols" of [r7rs](https://standards.scheme.org/official/r7rs.pdf), where it says "all symbols have the same names in the sense of `string=?`", so i just implemented them by simply implementing string equality for the symbols.

in section "6.1. Equivalence predicates" it states, for both `eqv?`:
> The `eqv?` procedure returns `#t` if:
>
> - [ ... ]
> - *obj1* and *obj2* are both symbols and are the same symbol according to the `symbol=?` procedure (section 6.5).

and for `eq?`:
> On symbols, [ ... ], `eq?` and `eqv?` are guaranteed to have the same behavior.

### DIFF
--- a/crates/steel-core/src/primitives.rs
+++ b/crates/steel-core/src/primitives.rs
@@ -58,7 +58,7 @@ use std::convert::TryFrom;
 use std::result;
 pub use streams::StreamOperations;
 pub use strings::string_module;
-pub use symbols::SymbolOperations;
+pub use symbols::symbol_module;
 pub use vectors::VectorOperations;
 
 macro_rules! try_from_impl {
--- a/crates/steel-core/src/primitives/symbols.rs
+++ b/crates/steel-core/src/primitives/symbols.rs
@@ -1,5 +1,53 @@
-use crate::rvals::{Result, SteelVal};
+use crate::rvals::{PrimitiveAsRef, RestArgsIter, Result, SteelString, SteelVal};
+use crate::steel_vm::builtin::BuiltInModule;
 use crate::stop;
+
+pub fn symbol_module() -> BuiltInModule {
+    let mut module = BuiltInModule::new("steel/symbols");
+    module
+        .register_value("concat-symbols", SymbolOperations::concat_symbols())
+        .register_value("symbol->string", SymbolOperations::symbol_to_string())
+        .register_native_fn_definition(SYMBOL_EQUALS_DEFINITION);
+    module
+}
+
+pub struct SteelSymbol<'a>(&'a SteelString);
+
+impl<'a> PrimitiveAsRef<'a> for SteelSymbol<'a> {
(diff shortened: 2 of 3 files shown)


mattwparas:
This is interesting, probably worth a discussion:

Steel symbols are by default interned if they are found at compile time. If they're generated at runtime, they won't be interned. This would generally mean that if you have some symbols that are statically written in a file, you should be able to run `eq?` against them and be confident that you're referring to only the one instance of that symbol.

So I think we don't want to change the default behavior of `ptr_eq` to use equality - that should still use the `Shared::ptr_eq` where relevant. Looking at Racket for example (which may not be abiding by the spec here):

```
~ ❯ racket                                                          
Welcome to Racket v8.15 [cs].
> (define foo 'foo)
> (define runtime-foo (string->uninterned-symbol "foo"))
> (eq? foo runtime-foo)
#f
> (eqv? foo runtime-foo)
#f
> (symbol=? foo runtime-foo)
#f
```

Looking at the spec again - it looks like there are lots of addendums for uninterned symbols:

```
Note: Some implementations have values known as “unin-
terned symbols,” which defeat write/read invariance, and also
violate the rule that two symbols are the same if and only if
their names are spelled the same. This report does not specify
the behavior of implementation-dependent extensions.

...

Note: The definition above assumes that none of the arguments
are uninterned symbols.
```

So maybe all we need to adjust for `symbol=?` is to also check `ptr_eq?` as well as the standard string equality? And then otherwise leave the `ptr_eq` implementation using the same behavior as before?

m4rch3n1ng:
i don't know enough about scheme to really "discuss" this with you (i am using steel as my interpreter of choice to learn this language), and i only originally opened this issue because the `symbol=?` procedure was missing so i'll just trust you on this one.

but i don't fully get what you want me to do for `symbol=?`? doing `ptr_eq? || string=?` should do the exact same as just doing `string=?` (albeit maybe sometimes slightly faster) and doing `ptr_eq? && string=?` should do the same as just doing `ptr_eq?`.

mattwparas:
Well, given you've taken the time to work on this, even if you don't (claim to) know enough about scheme to discuss, I still appreciate your input!

I do think you're right - I think to get the behavior that we want, we should just implement `symbol=?` via pointer equality, and not string equality.

m4rch3n1ng:
dropped the commit changing the behaviour of `eqv?` and changed `symbol=?` to compare the string pointers.

> Well, given you've taken the time to work on this, even if you don't (claim to) know enough about scheme to discuss, I still appreciate your input!

thanks i guess lol, but i haven't actually used symbols for pretty much anything apart from testing around with them while learning, realising i couldn't do `symbol=?` and then looking into r7rs if it would be doable implementing this myself.

mattwparas (approved):

PR #324, opened 2025-03-12
steel-language-server: fix offset for diagnostics with non-ascii utf8

m4rch3n1ng:
this fixes two related bugs regarding encoding:

1. steel spans work with utf-8 byte offsets (see [here](https://github.com/mattwparas/steel/blob/0a64032916f97f920679cab016fa4e965740ac74/crates/steel-parser/src/lexer.rs#L55) for example), but in the `offset_to_position` fn it used char offsets to properly calculate the line and line offsets.

2. the lsp protocol by default expects offsets to be in utf16 code units, but steel would previously give the offsets in utf8 code units (i.e. bytes).

it might be worth in the long run to set the encoding kind to utf8 (but `tower-lsp` currently doesn't seem to support that), but i think it's fine, especially 'cause utf16 has to be supported by all lsp clients.

before:

image
image

after:

image
image

### DIFF
--- a/crates/steel-language-server/src/backend.rs
+++ b/crates/steel-language-server/src/backend.rs
@@ -992,9 +992,15 @@ fn configure_lints() -> std::result::Result<UserDefinedLintEngine, Box<dyn Error
 }
 
 pub fn offset_to_position(offset: usize, rope: &Rope) -> Option<Position> {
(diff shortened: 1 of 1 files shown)


mattwparas:
This is a great catch, thank you!

mattwparas (approved):

PR #327, opened 2025-03-17
properly convert from lsp positions to byte offsets

m4rch3n1ng:
follow-up to #324

fixes some more cases of incorrect conversions between the utf16 the lsp protocol sends and the utf8 that steel uses. i already fixed the other way around in #324, but i somehow forgot that there are also conversions happening in the other direction: from lsp positions to utf8 bytes. this fixes those.

the position_to_offset function is not the most efficient thing i have ever written, but i don't think i can do much better since [ropey doesn't provide a utf16_cu_to_byte fn](https://docs.rs/ropey/latest/ropey/?search=utf16_cu_to_byte), and utf16 code units to utf8 bytes conversion is a little bit of a pain in the ass.

if a file contained a bunch of non ascii utf-8, then previously you could `hover` out of bounds of the actual identifier and would not be able to `hover` at the start of an identifier (hard to screenshot something not working so you just gotta trust me on this one):

image

after this change you can no longer `hover` outside of the bounds (still hard to screenshot), but now you can `hover` at the very start of the bounds again:

image

### DIFF
--- a/crates/steel-language-server/src/backend.rs
+++ b/crates/steel-language-server/src/backend.rs
@@ -195,11 +195,10 @@ impl LanguageServer for Backend {
         let definition = || -> Option<Hover> {
             let uri = params.text_document_position_params.text_document.uri;
             let mut ast = self.ast_map.get_mut(uri.as_str())?;
+            let rope = self.document_map.get(uri.as_str())?;
 
             let position = params.text_document_position_params.position;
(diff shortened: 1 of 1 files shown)


mattwparas (approved):

mattwparas:
Thanks for this - I appreciate you stress testing the LSP like this!

PR #328, opened 2025-03-17
steel-language-server: disable markdown feature and fix doc extraction

m4rch3n1ng:
this fixes two seperate, but slightly related issues:

1. the steel-language-server depended on `steel-core` with the `markdown` feature enabled, causing the builtin docs to be rendered, instead of being passed along as the raw markdown.

i'm not super happy with this being dependant on a feature instead of there being a seperate function to collect the docs as just strings, but i didn't want to duplicate that much code for something that is already fairly trivially possible, so i just left it.

2. when collecting doc comment lines, steel previously trimmed all whitespace at the start. this caused code blocks with indentation to break, as the indentation was stripped as well. additionally steel did not attempt to trim the trailing newlines, and then later joined all the comment lines with a "\n", resulting in the newlines getting duplicated.

before:
image
image

after:
image
image

### DIFF
--- a/crates/steel-language-server/Cargo.toml
+++ b/crates/steel-language-server/Cargo.toml
@@ -14,5 +14,5 @@ tower-lsp = { version = "0.20.0", features = ["proposed"] }
 serde = { version = "1.0.152", features = ["derive"] }
 dashmap = "5.1.0"
 log = "0.4.17"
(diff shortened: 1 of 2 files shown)


mattwparas (approved):

mattwparas:
Thank you!

PR #329, opened 2025-03-17
make expt calculations with ints higher precision

m4rch3n1ng:
make expt calculations use bigints and rationals instead of falling back to floats when possible.
specifically changes `(expt IntV IntV)` `(expt IntV BigInt)` `(expt BigInt IntV)` `(expt BigInt BigInt)`.

similar to the `/` procedure, only check for a negative denominator for `IntV`. 

before:
image

after:
image


mattwparas (approved):

mattwparas:
Thanks!

PR #331, opened 2025-03-18
parse more valid numbers

m4rch3n1ng:
there were a few numbers that were valid in scheme, but were not being parsed correctly by steel. i initally noticed this when i wanted to type `2e-4`, but that didn't work.

an incomplete list of numbers that are now supported:
- `2e-4` (which evaluates to `0.0002`)
- `2e+4` (which evaluates to `20000`)
- `3.0+inf.0i`
- `2e+4+inf.0i` (which evaluates to `20000+inf.0i`)
- `-inf.0i-2e-4` (which evaluates to `-0.0002-inf.0i`)
- `+nan.0+5.0i`
- `+inf.0+nan.0i`

image

### DIFF
--- a/crates/steel-core/src/rvals.rs
+++ b/crates/steel-core/src/rvals.rs
@@ -1356,6 +1356,14 @@ impl SteelComplex {
             _ => unreachable!(),
         }
     }
+
+    fn imaginary_is_finite(&self) -> bool {
+        match &self.im {
+            NumV(x) => x.is_finite(),
+            IntV(_) | Rational(_) | BigNum(_) | SteelVal::BigRational(_) => true,
+            _ => unreachable!(),
+        }
+    }
 }
 
 impl IntoSteelVal for SteelComplex {
@@ -1371,7 +1379,7 @@ impl IntoSteelVal for SteelComplex {
 
 impl fmt::Display for SteelComplex {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.imaginary_is_negative() {
+        if self.imaginary_is_negative() || !self.imaginary_is_finite() {
             write!(f, "{re}{im}i", re = self.re, im = self.im)
         } else {
             write!(f, "{re}+{im}i", re = self.re, im = self.im)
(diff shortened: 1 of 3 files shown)


mattwparas:
This looks great, thanks! I only have one reservation:

`-inf.0i-2e-4` doesn't seem to parse on Racket or Guile - not that its necessarily a deal breaker, but I'm curious why it doesn't parse there.

m4rch3n1ng:
oh yeah that is a little strange, especially cause `-2e-4-inf.0i` does seem to parse. i can look at the racket source code if you want, but i already tried that once and i find it very hard to read personally.

mattwparas:
No need to read the source, I'll poke around for a bit and see if I can get somewhere

m4rch3n1ng:
it seems like racket doesn't support `bi+a` syntax and only supports `a+bi`, but steel (built from master, not this branch) does support it:

image

r7rs does not mention anything about that syntax and only makes mention to `a+bi`

image

mattwparas:
Interesting - I think in general just supporting `a+bi` for consistency would make sense. It appears racket is very flexible with the definition of an identifier:

```
Welcome to Racket v8.15 [cs].
> (define 2i+2 10)
> 2i+2
10
```

So they probably make the distinction that only `a+bi` gets parsed into a number, and any other form would get interpreted as an identifier, such that there isn't any ambiguity.

I do see that steel does not make such a distinction, although I think it probably should follow the same form as Racket/Guile where only the `a+bi` form is parsed that way.

m4rch3n1ng:
do you want me to change it so that only `a+bi` syntax is parsed?

mattwparas:
If you are happy/willing to make that change, sure - otherwise, we can take this PR as is and then that component of the parsing can be adjusted later since it matches the existing behavior already.

Up to you

m4rch3n1ng:
doing that was a single line diff (+ updating the tests), but i also took the liberty to remove some unnecessary unicode decoding and replace two collects with the `smallvec!` macro.

mattwparas (approved):

mattwparas:
Thanks for this, appreciate it!

PR #333, opened 2025-03-21
implement more port procedures and move them to a builtin `steel/ports` module

m4rch3n1ng:
implements `call-with-port` `call-with-output-file` `call-with-input-file` `with-output-to-file` and `with-input-from-file` [r7rs](https://standards.scheme.org/official/r7rs.pdf) procedures.

~~since the `parameters.scm` file apparently not have access to the filesystem, i added those as a seperate prelude module "#%private/steel/ports" and then moved the "Port functions" from `parameters.scm` to the new modules.~~ ok apparently this still fails to build when trying to run something sandboxed. no idea how to fix that though, sorry. ~~it should probably work to add those functions to `stdlib.scm`? is that what i should do?~~ maybe i am just not very smart but i can't get those functions into `stdlib.scm` either. the only way i can think of would be to add those procedures as like rust impls, but that seems like it should be super unnecessary, but i don't know how to create a scm module and only include that when not sandboxed...

additionally i also implemented the `close-port` procedure.

i didn't add any tests, because i was unsure of how to test procedures that access the file system and i couldn't find tests for the already existing port procedures, but if you want me to i can add them.


mattwparas:
Let me poke around at your branch - the sandboxing stuff is a little mind bendy. Overall I like the change, so let's see what we can do to get this working

mattwparas:
Okay awesome, here is what we can do:

In `src/primitives/ports.rs` add the following functions:

```rust

#[function(name = "open-input-file")]
pub fn open_input_file_sandbox(path: &SteelString) -> Result<SteelVal> {
    stop!(Generic => "A sandboxed engine cannot open the file system.")
}

#[function(name = "open-output-file")]
pub fn open_output_file_sandbox(_: &SteelString) -> Result<SteelVal> {
    stop!(Generic => "A sandboxed engine cannot open the file system.")
}
```

And then add those as definitions to the module defined in `port_module_without_filesystem` within the same file.

That should get the builds passing

m4rch3n1ng:
oh yeah that makes a lot of sense, give me a minute and i'll implement them like that... thanks for the hint!

mattwparas:
Looks like the code coverage stuff is having issues (unrelated to your changes) - so ignore that failing on the build

mattwparas (approved):

mattwparas:
Thank you!

PR #336, opened 2025-03-26
Added docs to vectors.rs

gold-silver-copper:
Moved everything out of VectorOperations.
Added docs to all functions in vectors.rs which were registered.
Touched hashmaps.rs and transducers.rs to remove reference to VectorOperations.
Generated docs with 'cargo xtask docgen'

Thanks :)

### DIFF
--- a/crates/steel-core/src/primitives/vectors.rs
+++ b/crates/steel-core/src/primitives/vectors.rs
@@ -1240,6 +1240,18 @@ pub fn vec_ref(args: &[SteelVal]) -> Result<SteelVal> {
     }
 }
 
+/// Appends an element to the given vector.
+///
+/// (push elem vec) -> immutable-vector?
+///
+/// * elem : any - The element to append.
+/// * vec : immutable-vector? - The vector to which the element will be appended.
+///
+/// # Examples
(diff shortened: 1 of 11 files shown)


crates/steel-core/src/primitives/vectors.rs
```diff
@@ -263,7 +365,7 @@ fn immutable_vector_push(vector: &mut SteelVal, value: SteelVal) -> Result<Steel
     }
 }
 
-/// Pushes a value to the back of the vector, returning a new vector.
+// TODO: Register this function?
 #[steel_derive::function(name = "vector-push")]
 fn vector_push(vector: &mut SteelVal, value: SteelVal) -> Result<SteelVal> {
```
mattwparas, line 370: I take it this was never actually registered anywhere?

mattwparas: yeah a quick grep shows that it wasn't in use - it _probably_ should be registered though, but let me poke around a bit for why it wasn't originally

mattwparas (approved):

mattwparas:
Thank you!

PR #338, opened 2025-03-28
clean up dependency tree

m4rch3n1ng:
this pull request is purely a clean up for your dependency tree and should not change any of the features. i partially used [cargo-machete](https://github.com/bnjbvr/cargo-machete) for the low-hanging fruit (which is not super accurate) or just looked at the output from `cargo-tree` and kind of guessed and then looked if i can remove any of the dependencies.

1. ~~[64364e1e00ca897fc1c63d5b29caa68588dea3b4]: the `rustyline` crate by default enables a feature for custom keybinds (`custom-bindings`) which steel doesn't actually need and this was pulling in a few unnecessary dependencies~~ this is actually no longer possible since #349. whoops.
2. [3bd2990133c0d6aae53bfcb467526c34fec33fc5]: steel-core: remove dependency on `serde_derive` as you were already depending `serde` with the `derive` feature, and remove the `pretty` dependency. this doesn't actually change much since `serde` depends on `serde_derive` and `steel-parser` depends on `pretty`, but i think this makes the dependency graph a little cleaner.
3. [a11af18608be64a8dd540e3f798d0e5a19560012]: remove the `dirs` and `home` dependency from `steel-core`. steel was already pulling in the `env_home` crate through `which` and the `dirs` dependency was actually only used for getting the user home directory, which `env_home` can do too. if your msrv ever goes above 1.87 (the current nightly) you can actually just use `std::env::home_dir()` as that was very recently undeprecated (https://github.com/rust-lang/rust/pull/137327).
4. [350d6e5a75fc759329b7df542da303b7656e18e9]: similar to point 2 about `serde_derive`, just in all the other crates.
5. [68e3d64b9b4fef19ff289ab65ae719d9dc757f03]: remove dev-dependency on `serde` from steel-interpreter. as it was a dev dependency, this doesn't really affect anything, but it wasn't needed.
6. [fed7492d99d0092c86eb35a9c8319e9eac181c18]: i cannot fully replace `once_cell` with the `std` equivalents, as [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) was stabilized only in 1.80 and steel's msrv is 1.76, but i can replace it for the `steel-language-server`.
7. [371eaa62b82146fb9e9867241929db64d7f2414a]: update the `colored` crate to get rid of `lazy_static`. since `colored` was only used by the `steel-repl` which has an msrv of 1.81 anyway due to rustyline, it should be fine that colored v3 has an msrv of 1.80.
8. [2500789a95bac129809cf81c0ad3c4bad9c4f5d4]: `steel-core` was dependending on the `http` crate, but wasn't actually using it.
9. [3fcffd4882a06c4c5cadb5aaa99336043450b846]: there was already a comment in one of the `Cargo.toml`s about not pulling in the entirety of `num` and only depending on what was actually needed. this commit does exactly that: it replaces the `num` dependency and directly relies on `num-traits`, `num-rational` and `num-bigint`
10. [3c090a6aeb0d2fbf49e70ec57e0eedd4fe4e57d6]: similar to the commit with the `num` dependency, but instead with `crossbeam` where you only needed `crossbeam-channel` and `crossbeam-utils`. this change doesn't actually fully remove `crossbeam` from the dependency tree, as it is still used by `termimad`, but i opened a pr there as well: https://github.com/Canop/termimad/pull/69.
11. [75b6bbe4c6cadc649268ab8ea82e0cad7aea9c89]: remove unused `env_logger` dev dependency in steel-core.

none of these changes are "make-or-break" to me and i would be willing to drop any of the changes if you prefer not to have them.


mattwparas:
Thanks for this - overall looks good.

At some point in time I had left `once_cell` in because I was trying to keep the MSRV to match helix at 1.76, however it appears that my branch moved up to 1.81 (not sure when I did that, but that's fine) - I'm wondering if that is the only thing keeping the MSRV at 1.81? I'll do a little poking around before I make a decision here

m4rch3n1ng:
the 1.81 msrv is mostly due to the `home` crate, which raises msrv quite frequently. if you want, i can drop the commit replacing the `once_cell` dependency and replace your dependency on the `home` crate with `dirs`, which i think should allow me to drop the msrv of `steel-core` back down to 1.76, if you prefer. (`steel-repl` would still be msrv of 1.81 though, due to a transitive dependency)

m4rch3n1ng:
one more note: i realized that `owo-colors` is not actually a drop-in for `colored`, as colored also only colorizes if the terminal supports it, which would #312 for example not trivial to implement. *but* there actually is a new version of colored that gets rid of lazy_static anyway, and since i cannot bring the msrv of steel-repl under 1.81, it doesn't actually matter that colored 3 has an msrv of 1.80

mattwparas:
Thanks for the investigation there. I think, if possible, I'd like to try to keep the MSRV for steel core on 1.76 _if possible_. If its not trivially fixable with the home changes you've already made + removing the commit for the the once cell stuff, then we can merge this as is.

m4rch3n1ng:
done now. i used the `env_home` crate which i think should be good enough? (and it's already in your dependency tree anyway). the msrv goes back up to 1.81 if you enable the git feature, but i don't think i can fix that.

mattwparas:
I'm happy with this as is - would you be willing to fix the merge conflicts? If not, I can fix them :smiley:

m4rch3n1ng:
rebased and fixed the merge conflicts

m4rch3n1ng:
rebased again and fixed the other merge conflicts from #360

mattwparas:
agh, another PR just introduced a small merge conflict

m4rch3n1ng:
done again. also don't worry about it, my fault for doing a pretty big pr and i'd rather me do the merge conflicts than have other people have to worry about it (also most of the merge conflicts here were introduced by other prs from me)

mattwparas (approved):

PR #344, opened 2025-04-03
better parsing of numbers with radixes

m4rch3n1ng:
move radix parsing into the `parse_number` function in the lexer instead of handling it seperately. this makes it a lot easier to implement for example complex numbers and fractionals with radixes.

i couldn't find a way to parse floats with radixes, the closest i could get was the [lexical crate](https://crates.io/crates/lexical), but that only supports passing the number format as const generics if i'm not mistaken. i'll look another time, but for now they just aren't supported.

before:
image

after:
image


mattwparas (approved):

mattwparas:
Thanks!

PR #345, opened 2025-04-03
support complex numbers in polar notation

m4rch3n1ng:
r7rs specifies polar numbers in the $r@\theta$ notation where $r$ is the magnitude and $\theta$ is the angle in radians.

i haven't done extremely intensive testing as it is quite difficult to test as polar notation relies on fractions of pi and is therefore extremely imprecise.

while i was at it i also implemented the `make-rectangular` (trivial) `make-polar` (most of what this pr is) and the `angle` procedures.

a simple demo:
let $z := \frac{1}{2}+\frac{\sqrt{3}}{2}\mathrm{i}$. then $r := |z| = \sqrt{ \frac{1}{4} + \frac{3}{4} } = 1$, and $\theta = \cos^{-1}( \frac{1}{2} ) = \frac{\pi}{3}$, so $z = 1 \cdot ( \cos( \frac{\pi}{3} ) + \mathrm{i} \sin( \frac{\pi}{3} ) )$. since steel doesn't define a $\pi$ for me to use i will instead use `(acos -1)`

image

let $z := 8-8\sqrt{3}\mathrm{i}$. then $r = \sqrt{8^2 + (8 \cdot \sqrt{3})^2}=\sqrt{64+192}=16$, and $\theta=-\cos^{-1}(\frac{8}{16})=-\frac{\pi}{3}$. note the $-$, as $\mathrm{Im}(z)<0$. then $z = 16 \cdot ( \cos( -\frac{\pi}{3} ) + \mathrm{i}\sin( -\frac{\pi}{3} ) )$

image


mattwparas (approved):

mattwparas:
thanks!

PR #347, opened 2025-04-04
allow range builtin to be called with only one argument

m4rch3n1ng:
allow the `(range)` function to only take 1 argument. this makes it a `0..m` range. this also fixes the docs, as they confused the `(` and `[` brackets (i think lol).

i was not sure how to document functions that take one or two arguments, so i took inspiration from racket and documented both options seperately.

i was also not sure which arity to use, as `Range(n)` only takes one number and i would need a `Range(1, 2)` (i think), so i just picked `AtMost(2)`.

### DIFF
--- a/crates/steel-core/src/primitives/lists.rs
+++ b/crates/steel-core/src/primitives/lists.rs
@@ -295,31 +295,36 @@ macro_rules! debug_unreachable {
 //     }
 // }
 
-/// Returns a newly allocated list of the elements in the range (n, m]
+/// Returns a newly allocated list of the elements in the range [n, m) or [0, m) when n is not given.
 ///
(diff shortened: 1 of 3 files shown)


gold-silver-copper:
I could try implementing a range(n,m) arity check and the proper derive code for it if @mattwparas wants it to be like that, rather than taking a single value as it currently does.

mattwparas:
@gold-silver-copper - yeah, I think the spirit of the range was to have a lower and upper bound where the lower bound was non-zero. Probably just got forgotten along the way. If you wanted to go ahead and implement that, it would be appreciated :slightly_smiling_face:

mattwparas (approved):

PR #348, opened 2025-04-04
improve transducers

m4rch3n1ng:
the biggest change here is that i implemented the `execute` iterator consumer. the possible return types specified by the symbol are `'list`, `'vector`, `'string`, `'hashset` and `'hashmap`. there is a few things that i was not quite sure with how to do so i just did the best i could. when executing on a `MutableVector` it currently collects into a `VectorV` and when executing on a `StreamV` it currently returns a `ListV`. i didn't want to implement another reducer for this, but i think that that will be doable another time.

the comment about the `Range(n)` arity from #347 also applies here.

as i am not very good at writing documentation, the documentation is a little basic, but i think the examples mostly speak for themselves.

i also implemented VmCore::previous_span. for execute i needed a steel_derive::context() and i wanted to give useful errors, and this was the cleanest solution that i managed to come up with.

### DIFF
--- a/crates/steel-core/src/primitives/transducers.rs
+++ b/crates/steel-core/src/primitives/transducers.rs
@@ -1,8 +1,11 @@
 use crate::gc::Gc;
 use crate::rvals::SteelVal::*;
 use crate::rvals::{Result, SteelVal};
-use crate::steel_vm::builtin::BuiltInModule;
-use crate::stop;
+use crate::steel_vm::{
+    builtin::BuiltInModule,
+    vm::{VmContext, VmCore},
+};
+use crate::{builtin_stop, stop};
 
 use crate::values::transducers::Transducer;
 use crate::values::transducers::Transducers;
@@ -26,6 +29,7 @@ pub fn transducer_module() -> BuiltInModule {
 
     module
         .register_native_fn_definition(COMPOSE_DEFINITION)
+        .register_native_fn_definition(TRANSDUCE_DEFINITION)
         .register_native_fn_definition(MAPPING_DEFINITION)
         .register_native_fn_definition(FLATTENING_DEFINITION)
         .register_native_fn_definition(FLAT_MAPPING_DEFINITION)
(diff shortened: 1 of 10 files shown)


crates/steel-core/src/primitives/transducers.rs
```diff
@@ -33,25 +81,103 @@ pub fn compose(args: &[SteelVal]) -> Result<SteelVal> {
...
+    let symbol = match args.get(2) {
+        Some(SymbolV(symbol)) => Some(symbol),
+        Some(invalid) => {
+            builtin_stop!(TypeMismatch => format!("third argument should be a symbol, got {}", invalid); ctx.previous_span())
+        }
+        None => None,
+    };
 
+    let reducer = if let Some(symbol) = symbol {
+        match &****symbol {
```
mattwparas, line 122: What is going on here with all the de-referencing?

mattwparas:
What was the motivation behind the `execute` function compared to explicitly providing the reducer via something like `(into-list)`?

crates/steel-core/src/primitives/transducers.rs
m4rch3n1ng: it derefs `&SteelString` => `SteelString` => `Gc<String>` => `String` => `str`.

m4rch3n1ng:
it is mentioned in the [steel book: chapter transducers](https://mattwparas.github.io/steel/book/reference/transducers.html), and i modeled it after that, because when i wanted to try it, it didn't exist.

m4rch3n1ng:
i am not entirely sure what you mean by "merging [my] work into [that] function". do you mean making the reducer optional and just default-transducing it to itself? i don't think that would be a great api to use. i thought `execute` 

i would also personally advocate for swapping the order of arguments for `transduce`: `execute` takes the collection as the last argument, but transduce takes it as the first. i personally think it makes more sense for the list to be the last argument (and clojure agrees, their reduce function also takes the 

if you want me to i can drop the `execute` function and update the docs, but i personally thought that the function makes a lot of sense (and is pretty clean in the examples) and thought of it akin to the rust `collect`: sure you can also do collect via `reduce(Vec::new(), || vec.push)`, but having a collect that just "does that for you" would be nice.

also i personally thought it was confusing to

mattwparas:
Sorry I suppose you're right with the merging the work together. I took another look and they're fairly distinct, so that is fine. Apologies for the confusion.

`transduce` is fairly ubiquitous and I don't think changing the order of the arguments makes a lot of sense given that it would be a breaking change. The api I think is as of now:

```scheme
(transduce iterable transducer ... reducer)
```

Given this, I don't think an `execute` function is explicitly necessary on its own, besides being a slightly different expression of the same api. `execute` looks like:

```scheme
(execute iterator? iterable) -> any/c
(execute iterator? iterable symbol?) -> any/c
```

Can't one be expressed in terms of the other, and vice versa? I think I was inspired by a Racket library for transducers when formulating this version of the function, which would explain the difference between clojure and this.

m4rch3n1ng:
ok that is fair if you want to keep that in one function.

i think i found the library you mentioned (https://docs.racket-lang.org/rebellion/Transducers.html) and it's transduce function has the arguments in the order that you have currently, instead of what clojure does.
thinking it over i do think it makes sense to "keep it all in one place", especially since the `transduce` function is a lot more explicit as you have to explicitly declare what you want to transduce into.

(one last thing to note though is that the racket library takes the reducer as just `into-list` instead of having to be called as `(into-list)`. i think that would make sense to implement here, though it would be a major breaking change as well.)

i'll give this a second shot some time soon and then just update the docs and throw out the `execute` function, though you'll have to give me a little time here as well ...

m4rch3n1ng:
i've removed all mentions of the `execute` functions and the examples and the refernce now use the `transduce` function instead. the reference is a little shorter now, as i didn't add more info and mainly just removed mentions of `execute`, as i am very bad at writing docs...

mattwparas (approved):

mattwparas:
Thanks for this. Apologies for the delay.

m4rch3n1ng:
no worries, the delay was mostly on me anyway. thanks for the merge!

PR #360, opened 2025-04-11
properly implement the `string->number` procedure

m4rch3n1ng:
instead of essentially calling `eval`, just call the `parse_number` in `steel-parser::lexer`, with an optional radix. this also has the upside of fixing most of the parsing inconsistencies essentially for free.

before:
image

after:
image

this still leaves `number->string` in need of a serious overhaul, but i've been working on that too.


mattwparas (approved):

mattwparas:
Thanks!

PR #361, opened 2025-04-11
support parsing numbers without the integer part

m4rch3n1ng:
this basically just means parsing `.2` as `0.2`. this is trivial, as the rust `FromStr` impl for `f64` already supports it out of the box. it was also already partially supported, as `2+.2i` parsed just fine, and the imaginary part of that is just `.2`.

specified in [r7rs](https://standards.scheme.org/official/r7rs.pdf) on page 62 in the $\<\text{decimal 10}>$ rule.


mattwparas (approved):

mattwparas:
Thanks!

PR #368, opened 2025-04-17
Fix dylib windows support

ck961018:
When attempting to use dylib on Windows platforms, the `cargo steel-lib` command incorrectly copies incompatible dependency libraries:

企业微信截图_1744873444840

Even after manually placing the correct libraries in the native directory, Steel still fails to recognize them:

企业微信截图_17448734131838

So I try to Identify and modify the related code. The changes was tested on Win11 and WSL2(Arch Linux).


mattwparas (approved):

mattwparas:
Thank you!

PR #371, opened 2025-04-20
Implement proc macro for quasiquoting

mattwparas:
Example usage:

```rust
#[test]
fn test_steel_quote_macro() {
    let foobarbaz = ExprKind::atom("foo");
    let foobarbaz_list = ExprKind::List(List::new(vec![ExprKind::atom("foo")]));

    let expanded = steel_derive::internal_steel_quote! {
        (define bananas #foobarbaz)
        (define x (begin @foobarbaz_list ...))
    }
    .unwrap();

    for expr in expanded {
        println!("{}", expr);
    }
}

```

Prints:

```
(define bananas foo)
(define x (begin foo))
```

### DIFF
--- a/crates/steel-core/src/parser/replace_idents.rs
+++ b/crates/steel-core/src/parser/replace_idents.rs
@@ -40,6 +40,25 @@ pub fn replace_identifiers(
     ReplaceExpressions::new(bindings, binding_kind, fallback_bindings).visit(expr)
 }
 
+pub fn expand_template_pair(
+    mut exprs: Vec<ExprKind>,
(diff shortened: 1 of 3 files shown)


mitchmindtree:
Lgtm! Though I wonder if it should be just `steel_quote!` now that it's no longer just internal? Also might be handy to have it re-exported from `steel` too?

mattwparas:
> Lgtm! Though I wonder if it should be just `steel_quote!` now that it's no longer just internal? Also might be handy to have it re-exported from `steel` too?

Should be good now - re exported the `steel_quote` - there are separate implementations for the internal one so I'm keeping that around

PR #379, opened 2025-04-21
add repl builder

### LINKED ISSUE #376: Custom REPL display_startup
Hey, this is a fairly simple feature request: can we have a customizable REPL startup message? The use case is an embedded REPL, and we could take advantage of that message to provide some sort of info relevant to the domain, like which symbols are exported from host application and could be accessed.

I don't mind implementing it if you don't have anything against it, also I would like to hear if there is something else worth adding to the REPL-builder structure besides custom welcome message

fominok:
This shall resolve #376 by adding an optional builder to provide custom welcome message as well as configurable history path


fominok:
I'm not quite sure about that coverage thing that fails, like what can I possibly test there that would make sense in putting effort into covering it

mattwparas:
Don't worry about the coverage, I'll take a look at this PR after work

crates/steel-repl/src/repl.rs
```diff
@@ -213,161 +208,209 @@ impl ConditionalEventHandler for CtrlCHandler {
...
+            startup_text: None,
+            history_path: None,
+        }
+    }
+}
 
-    #[cfg(not(target_os = "windows"))]
-    let mut prompt = format!("{}", "λ > ".bright_green().bold().italic());
+impl<S: Display, P: AsRef<Path> + Debug> Repl<S, P> {
+    pub fn with_startup<NS: Display>(self, startup_text: NS) -> Repl<NS, P> {
```
mattwparas, line 228: Do you need the generics here? Can you just reference the `S` from the impl?

Same for the `with_history_path`

fominok: That's what I did at first, but then if some builder methods aren't used then some type parameters can't be inferred, so `new` specifies defaults and then separate generics allow to override those defaults 

I tried to use syntax for default generic parameters but had no luck in this case

mattwparas (approved):

mattwparas:
ah looks to be some merge conflicts, otherwise LGTM

fominok:
@mattwparas I've updated the PR with recent changes, should be good now; I suppose those failing checks don't allow me to merge even with approval, so I'm leaving it to you

PR #380, opened 2025-04-24
properly implement `number->string`

m4rch3n1ng:
follow-up to #360. there were a few issues previously with the `number->string` procedure:

- it could not convert rationals and complex numbers
- bigints were always printed with the radix 10, even if a radix was given
- printing negative numbers with a radix would print them as their two's complement

i have fixed all of these in this pr.

before:
image

after:
image

### DIFF
--- a/crates/steel-core/src/primitives/strings.rs
+++ b/crates/steel-core/src/primitives/strings.rs
@@ -5,6 +5,8 @@ use crate::rvals::{IntoSteelVal, RestArgsIter, Result, SteelByteVector, SteelStr
 use crate::steel_vm::builtin::BuiltInModule;
 use crate::{stop, Vector};
 
+use std::fmt::Write;
+
 use steel_derive::{function, native};
 
 /// Strings in Steel are immutable, fixed length arrays of characters. They are heap allocated, and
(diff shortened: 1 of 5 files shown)


crates/steel-core/src/primitives/strings.rs
```diff
@@ -107,21 +109,100 @@ pub fn char_equals(rest: RestArgsIter<char>) -> Result<SteelVal> {
...
     }
+
+    Ok(())
+}
+
+fn number_to_string_impl(value: &SteelVal, radix: Option<usize>) -> Result<SteelVal> {
+    let mut accumulator = Vec::new();
+    format_number(&mut accumulator, value, radix.filter(|x| *x != 10))?;
+
+    let string = String::from_utf8(accumulator).expect("should just be ascii");
```
m4rch3n1ng, line 204: this could be `let string = unsafe { String::from_utf8_unchecked(accumulator) };`, as i can guarantee that i am only pusing valid utf8 (ascii) to the vec, but i didn't want to add any unsafe without your feedback.

the num-bigint function that i use for formatting bigints [also does it with unchecked](https://github.com/rust-num/num-bigint/blob/a25836ec6c341d1aa40c97335842f330b6a62911/src/bigint.rs#L880)

crates/steel-core/src/primitives/strings.rs
```diff
@@ -107,21 +109,100 @@ pub fn char_equals(rest: RestArgsIter<char>) -> Result<SteelVal> {
...
+        }
+        if value < 0 {
+            acc.push(b'-');
+        }
+
+        acc[start..].reverse();
+    }
+
+    pub fn big(acc: &mut Vec<u8>, value: BigInt, radix: usize) {
+        let fmt = value.to_str_radix(radix as u32);
```
m4rch3n1ng, line 137: not entirely happy with this allocation, but i didn't want to reinvent the wheel by implementing it myself (and the num-bigint crate itself allocates [in its `Display` impl](https://github.com/rust-num/num-bigint/blob/a25836ec6c341d1aa40c97335842f330b6a62911/src/bigint.rs#L140-L144))

crates/steel-core/src/primitives/strings.rs
mattwparas: I think for now I'd prefer not adding any unsafe - if you want to slap it behind a feature flag that would be fine, but for now this is okay

mattwparas (approved):

PR #381, opened 2025-04-25
implement `inexact` from r7rs

m4rch3n1ng:
r7rs defines the new procedures `exact` and `inexact` to replace the r5rs `inexact->exact` and `exact->inexact` procedures, respectively. i don't think it makes sense to remove the old names, but i do think it makes sense to create aliases for them.

previously steel implemented `inexact->exact`, `exact->inexact` and `exact`. the previous `exact` and `inexact->exact` procedures were implemented seperately and behaved very differently. this merges them two one procedure and essentially just creates an alias. it also creates an alias for `inexact` that points to `exact->inexact`

also i fixed an issue where `inexact->exact` (and `exact`) returned `#false`, when trying to exactify `nan` or `inf`. now it raises an error.


mattwparas (approved):

mattwparas:
thank you!

PR #404, opened 2025-05-26
optimize a few allocations and clones out of the transducer

m4rch3n1ng:
previously the transduce function would , while it would be possible to do this with only one. this is tackled in [1a041bdd494b017141c31cdcdeb4c337dca25d27] with a `FlattenOk` enum, that either returns the `Err` or flattens the value of `Ok` into multiple `Ok`s.

the `FlattenOk` enum dispatcher was the idea of [Itertools::flatten_ok](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.flatten_ok).

in [086faf26fdae690e44a7b20b2752b9e6054c9dfb] i save an `unwrap` (a clone out of a ref-count) and just use a reference.

both of these have fixed a `TODO` comment, which was the reason that i even started looking into these.

### DIFF
--- a/crates/steel-core/src/primitives/transducers.rs
+++ b/crates/steel-core/src/primitives/transducers.rs
@@ -5,10 +5,8 @@ use crate::steel_vm::{
     builtin::BuiltInModule,
     vm::{VmContext, VmCore},
 };
-use crate::{builtin_stop, stop};
-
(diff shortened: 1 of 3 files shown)


mattwparas (approved):

mattwparas:
Thank you!

PR #405, opened 2025-05-28
optimize real_literal_to_steelval and deduplicate number_literal_to_steel

m4rch3n1ng:
the `real_literal_to_steel` function was 1) eagerly allocating when converting to a `Rational` and 2) was panicking when the denominator of the `Rational` was 0, instead of returning an error.

additionally the `number_literal_to_steel` function was essentially a reimplementation of `NumberLiteral::into_steelval`. i also changed it to an `impl IntoSteelval for &NumberLiteral`, as i think that makes sense.


mattwparas (approved):

mattwparas:
Thanks!

PR #440, opened 2025-06-29
Add docs for list->string, list-drop and list-tail

godalming123:


docs/src/builtins/steel_lists.md
```diff
@@ -269,10 +270,53 @@ error[E11]: Generic
...
+
+#### Examples
+```scheme
+(list->string '(#\a #\b #\c)) ;; => "abc"
+```
+
 ### **list->vector**
 ### **list-chunks**
 ### **list-drop**
+Remove a certain number of elements (`n`) from the front of `lst`.
```
mattwparas, line 287: can you add these docs on the functions themselves in `primitives/lists.rs`? This is where these get generated from.

PR #447, opened 2025-07-04
Several documentation improvements

godalming123:
- Fix unclosed code block in docs
- Add documentation for list->string, list-drop and list-tail
- Add contributor documentation on regenerating steel documentation


crates/steel-core/src/primitives/lists.rs
```diff
@@ -788,6 +819,16 @@ pub fn list_ref(list: &List<SteelVal>, index: isize) -> Result<SteelVal> {
         .ok_or_else(throw!(Generic => format!("out of bounds index in list-ref - list length: {}, index: {}", list.len(), index)))
 }
 
+/// Convert a list of charecters to a string.
```
godalming123, line 822: I'm not sure why this doc string isn't being generated into the `docs/` directory.

mattwparas (approved):

PR #459, opened 2025-07-25
fix new rustc lints

m4rch3n1ng:
rust beta (i.e. the next stable release) introduces two more rustc lints, mismatched-lifetime-syntaxes:

```
warning: hiding a lifetime that's elided elsewhere is confusing
   --> crates/steel-core/src/steel_vm/ffi.rs:176:21
    |
176 |     fn into_ffi_arg(&self) -> RResult<FFIArg, RBoxError>;
    |                     ^^^^^             ------ the same lifetime is hidden here
    |                     |
    |                     the lifetime is elided here
    |
    = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
    |
176 |     fn into_ffi_arg(&self) -> RResult<FFIArg<'_>, RBoxError>;
    |                                             ++++
```

for that lint, i just went through every single warning and just fixed them.

and unpredictable_function_pointer_comparisons:

```
warning: function pointer comparisons do not produce meaningful results since their addresses are not guaranteed to be unique
   --> crates/steel-core/src/steel_vm/builtin.rs:175:15
    |
173 | #[derive(Copy, Clone, Hash, PartialEq, Eq)]
    |                             --------- in this derive macro expansion
174 | pub enum BuiltInFunctionType {
175 |     Reference(FunctionSignature),
    |               ^^^^^^^^^^^^^^^^^
    |
    = note: the address of the same function can vary between different codegen units
    = note: furthermore, different functions could have the same address after being merged together
    = note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>
    = note: `#[warn(unpredictable_function_pointer_comparisons)]` on by default
```

there was not much i could do for that lint, so i just `#[allow]`ed it. the official suggestion is to be more explicit and use `std::ptr::fn_addr_eq` for more clarity, but i don't think manually writing the `PartialEq` impl to use that function is better. also `fn_addr_eq` is only available in rust >= 1.85.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2536,9 +2536,9 @@ dependencies = [
 
 [[package]]
 name = "onig_sys"
-version = "69.8.1"
+version = "69.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7b829e3d7e9cc74c7e315ee8edb185bf4190da5acde74afd7fc59c35b1f086e7"
+checksum = "c7f86c6eef3d6df15f23bcfb6af487cbd2fed4e5581d58d5bf1f5f8b7f6727dc"
 dependencies = [
  "cc",
  "pkg-config",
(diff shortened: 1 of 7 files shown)


mattwparas (approved):

mattwparas:
thanks!

PR #470, opened 2025-07-31
fix colorized output in display

jrvidal:
I think https://github.com/mattwparas/steel/pull/339 broke colorized output. Current behavior is:
```
(display #\u{1b})

=> #\u001b
```

which is incorrect, as `display` should not print external representations.


mattwparas (approved):

mattwparas:
Thank you!

PR #473, opened 2025-08-02
match number literals in macros by value

jrvidal:
I think integer literal matching can be simplified by just storing `NumberLiteral` in macro patterns and comparing "by value", a la `into_steelval()`. This allows things like:
```scheme
(define-syntax foo (syntax-rules ()
  [(_ 3/2) #t]))
=>

(foo 6/4)
=> #t
```


crates/steel-parser/src/lexer.rs
```diff
@@ -336,14 +337,14 @@ impl<'a> Lexer<'a> {
                     self.eat();
                 }
                 '(' | ')' | '[' | ']' => {
-                    return if let Some(t) = parse_number(self.slice(), None) {
+                    return if let Some(t) = try_parse_number(self.slice(), None)? {
```
jrvidal, line 340: :information_source: I've noticed that many implementations treat `1/0` literals as a _syntax_ error, instead of a runtime error. This does allow us to assume that `IntoSteelVal` for `NumberLiteral` is infallible.

crates/steel-parser/src/tokens.rs
```diff
@@ -92,8 +92,6 @@ pub enum TokenType<S> {
     Identifier(S),
     Keyword(S),
     Number(Box<NumberLiteral>),
-    // TODO: Consider using Arc<String> here instead to save
```
jrvidal, line 95: Maybe outdated?

jrvidal:
OK, fixed my conflicts with the `ManyConstant` stuff. I was actually planning to tackle this next :sweat_smile: I might look into refactoring all those `Many<Stuff>` into a single `Many(Box<MacroPattern>)`.

crates/steel-parser/src/tokens.rs
mattwparas: yeah that is outdated, good catch

crates/steel-parser/src/lexer.rs
mattwparas: Just so I'm following, it looks like you have this fail to parse the macro if it detects that? I don't think there is any handling in the parser for that, but that seems like a good idea otherwise

mattwparas (approved):

PR #488, opened 2025-08-27
 fix char display in the repl again 

m4rch3n1ng:
i "fixed" the repl char display in #339, which accidentally broke the `(display)` functions, which was then fixed in #470, ... which broke the char display in the repl again, as that was just calling `(displayln)` instead of the rust `Display` impl...

this contains some small, additional cleanups that i noticed while attempting to fix the char printing, specifically removing the `SteelValDisplay` from #470 and just moving that into the regular `Display` impl, while still having the `Debug` impl for how the repl and the `write` function wants the value.

here's a quick demo:

image

### DIFF
--- a/crates/steel-repl/src/repl.rs
+++ b/crates/steel-repl/src/repl.rs
@@ -82,14 +82,7 @@ fn finish_load_or_interrupt(vm: &mut Engine, exprs: String, path: PathBuf) {
     match res {
         Ok(r) => r.into_iter().for_each(|x| match x {
             SteelVal::Void => {}
-            SteelVal::StringV(s) => {
-                println!("{} {:?}", "=>".bright_blue().bold(), s);
-            }
-            _ => {
-                print!("{} ", "=>".bright_blue().bold());
-                vm.call_function_by_name_with_args("displayln", vec![x])
-                    .unwrap();
-            }
+            _ => println!("{} {}", "=>".bright_blue().bold(), x),
         }),
         Err(e) => {
             vm.raise_error(e);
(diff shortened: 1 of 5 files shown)


crates/steel-repl/src/repl.rs
```diff
@@ -82,14 +82,7 @@ fn finish_load_or_interrupt(vm: &mut Engine, exprs: String, path: PathBuf) {
...
         Ok(r) => r.into_iter().for_each(|x| match x {
             SteelVal::Void => {}
-            SteelVal::StringV(s) => {
-                println!("{} {:?}", "=>".bright_blue().bold(), s);
-            }
-            _ => {
-                print!("{} ", "=>".bright_blue().bold());
-                vm.call_function_by_name_with_args("displayln", vec![x])
-                    .unwrap();
-            }
```
mattwparas, line 92: Why is the call to `displayln` being removed? That implementation lives in `print.scm` and functions a little differently than the builtin Display implementation, because it handles how structs can be printed

m4rch3n1ng: sorry for the spam, but i also do not see where `display` and `Display` disagree about the printing of structs (the following is run on this branch):

image

mattwparas: I modeled off of this https://docs.racket-lang.org/reference/Printer_Extension.html?q=gen%3Acustom-write#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._gen~3acustom-write%29%29

But in general, here is an example in steel

```scheme
λ > (struct Foo () #:printer (lambda (obj printer) (simple-display "hello world")))
λ > (Foo)
=> hello world
λ > (displayln (Foo))
hello world
```

m4rch3n1ng: yeah i just read that part of the docs. in racket, even `write` calls the custom printer, which i don't know if that is the best behaviour? as `write` is specified to be "machine-readable".

i may have bitten off a little more than i can chew here, as i am not very sure how i would do that in scheme, as the display impl looks quite complicated, but i'll look into it!

m4rch3n1ng:
closing in favor of #489

PR #489, opened 2025-08-27
introduce the `(print)` procedure

m4rch3n1ng:
i wanted to just get the `(print)` procedure working for now, so i haven't fixed the `display` impl, if you want me to, i could add it here, or i would add it in a separate pr sometime in the future, whatever you prefer.

i also added a few tests in a new `steel-core/src/tests/success/print.scm`, though most of those are commented out, as i haven't fixed the `(display)` impl yet, so most of those should be more like an "ideal", which we are not yet at, but i think it is nice to have as a kind-of reference.

as a note, i made the `(println)` procedure (which i already configured to be used in the repl) more in line with the `print/display` impl, letting it take a port as the second argument. i was personally annoyed, that `displayln` did not take a port (as it doesn't in ), but if you prefer, i can do the same for `println`


crates/steel-core/src/scheme/print.scm
```diff
@@ -4,17 +4,175 @@
...
+                     (simple-display " . ")
+                     (#%print (cdr obj) collector)
+                     (simple-display ")"))
+                   (cdr list-obj)))
+       (simple-display ")"))]
+    [(#%private-struct? obj)
+     (let ([printer (#%struct-property-ref obj '#:printer)])
+       (if (function? printer)
+           (printer obj (λ (x) (#%print x collector)))
+           (simple-display obj)))]
```
m4rch3n1ng, line 174: in the `#%display` functions, there is another case here, when `printer` is not a function, but is truthy, where it prints `"#<" + #:name + ">"`, where i wasn't sure, why that case existed and nothing in steel passes anything to #:printer, that isn't a lambda, so i just left it and wanted to ask you now...

mattwparas (approved):

PR #492, opened 2025-08-29
rename `(read-line-from-port)` to `(read-line)`, trim trailing newline

m4rch3n1ng:
this makes it imo more in-line with the other port reading functions,
`read` is not named `read-from-port` for example, and it matches the name
of the equivalent r7rs procedure.

in addition, the r7rs spec says, that the return value is "a string containing all of
the text up to *(but not including)* the end of line".

the reason for the size of this pr is, that it changes the output of `tail.scm`, since that reads from a file via `read-line`, which results in half the newlines, which results in half the lines, which is a -30k diff ...

additionally it has apparently been a while since someone ran `cargo xtask docgen`, which is why there is a pretty large diff there as well. the actual change is less than 100 lines diff.

### DIFF
--- a/crates/steel-core/src/primitives/ports.rs
+++ b/crates/steel-core/src/primitives/ports.rs
@@ -40,7 +40,8 @@ pub fn port_module() -> BuiltInModule {
         .register_native_fn_definition(WRITE_CHAR_DEFINITION)
         .register_native_fn_definition(FLUSH_OUTPUT_PORT_DEFINITION)
         .register_native_fn_definition(READ_PORT_TO_STRING_DEFINITION)
-        .register_native_fn_definition(READ_LINE_TO_STRING_DEFINITION)
+        .register_native_fn_definition(READ_LINE_DEFINITION)
+        .register_native_fn_definition(READ_LINE_FROM_PORT_DEFINITION)
         .register_native_fn_definition(GET_OUTPUT_STRING_DEFINITION)
         .register_native_fn_definition(GET_OUTPUT_BYTEVECTOR_DEFINITION)
         .register_native_fn_definition(IS_INPUT_DEFINITION)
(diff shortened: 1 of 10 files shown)


crates/steel-core/src/primitives/ports.rs
```diff
@@ -390,20 +390,14 @@ pub fn is_output(maybe_port: &SteelVal) -> bool {
     }
 }
 
-#[function(name = "read-line-from-port")]
-pub fn read_line_to_string(port: &SteelPort) -> Result<SteelVal> {
-    let res = port.read_line();
+#[function(name = "read-line")]
```
mattwparas, line 393: Rather than rename this function, could we just make a new one that is `read-line`, and leave a comment on the `read-line-from-port` specifying that `read-line` is preferred? Otherwise this is a breaking change

crates/steel-core/src/values/port.rs
```diff
@@ -165,6 +165,17 @@ impl SteelPortRepr {
...
             // FIXME: fix this and the functions below
             _ => stop!(TypeMismatch => "expected an input port"),
+        };
+
+        if size == 0 {
+            Ok(None)
+        } else {
+            // r7rs read-line doesn't return the trailing \n
+            if string.ends_with('\n') {
+                string.truncate(string.len() - 1);
```
mattwparas, line 175: Could we do `string.pop()` here?

crates/steel-core/src/primitives/ports.rs
m4rch3n1ng: done, `read-line-from-port` still returns the newline and i reverted all previously made changes from `read-line-from-port` to `read-line`, and only kept the ones in r7rs and the steel examples.

mattwparas (approved):

PR #495, opened 2025-09-01
add a Display impl for ports

m4rch3n1ng:
before:

```
λ > (open-input-string "test")
=> #<port>
```

after:

```
λ > (open-input-string "test")
=> #<input-port:string>
```


m4rch3n1ng:
the macos failure seems spurious?

PR #499, opened 2025-09-07
better messages for tokenizer errors

jrvidal:
This adds more detail to our tokenizer errors. In particular, it sets correct errors spans so the error is pinpointed instead of reporting the whole token.

## Demo
```
error[E09]: Parse
  ┌─ ./steel/foo.scm:3:14
  │
3 │ "adsfadsf\   a
  │              ^ Parse: Syntax Error: unexpected character, expected whitespace or newline

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:10
  │
3 │ "adsfadsf\zxx"
  │          ^^ Parse: Syntax Error: invalid escape 'z'

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:5
  │
3 │ #(x #\hello z)
  │     ^^^^^^^ Parse: Syntax Error: invalid character name

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:10
  │
3 │ #(x "asdf\xdeadbeef" z)
  │          ^^^^^^^^^^ Parse: Syntax Error: unclosed hex escape, expected ';'

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:5
  │
3 │ #(x #\u{dead z)
  │     ^^^^^^^^ Parse: Syntax Error: unclosed hex escape, expected '}'

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:5
  │
3 │ #(x #\xafz z)
  │     ^^^^^^ Parse: Syntax Error: invalid hex escape literal, invalid digit found in string

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:9
  │
3 │ #(x "foo\u{d803}adsfa" z)
  │         ^^^^^^^^ Parse: Syntax Error: invalid code point d803

error[E09]: Parse
  ┌─ ./steel/foo.scm:3:9
  │
3 │ #(x "foo\u{deadbeefdeadbeef}adsfa" z)
  │         ^^^^^^^^^^^^^^^^^^^^ Parse: Syntax Error: invalid hex escape literal, number too large to fit in target type

```

### DIFF
--- a/crates/steel-parser/src/lexer.rs
+++ b/crates/steel-parser/src/lexer.rs
@@ -7,6 +7,7 @@ use smallvec::{smallvec, SmallVec};
 use std::borrow::Cow;
 use std::char;
 use std::iter::Iterator;
+use std::ops::Range;
 use std::sync::Arc;
 use std::{iter::Peekable, str::Chars};
 
@@ -38,6 +39,7 @@ pub struct Lexer<'a> {
     queued: Option<TokenType<InternedString>>,
     token_start: u32,
     token_end: u32,
+    error: Range<u32>,
 
     ident_buffer: String,
 }
@@ -50,6 +52,7 @@ impl<'a> Lexer<'a> {
             queued: None,
             token_start: 0,
             token_end: 0,
+            error: Default::default(),
             ident_buffer: String::new(),
         }
     }
@@ -87,7 +90,7 @@ impl<'a> Lexer<'a> {
             match c {
                 '"' => return Ok(TokenType::StringLiteral(Arc::new(buf))),
                 '\\' => {
-                    if let Some(c) = self.read_string_escape()? {
+                    if let Some(c) = self.read_string_escape(TokenError::IncompleteString, '"')? {
                         buf.push(c);
                     }
                 }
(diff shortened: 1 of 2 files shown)


crates/steel-repl/src/highlight.rs
```diff
@@ -129,7 +129,9 @@ impl Validator for RustylineHelper {
                 Err(err) => {
                     unfinished = std::matches!(
                         err.ty,
-                        TokenError::IncompleteString | TokenError::IncompleteIdentifier
+                        TokenError::IncompleteString
+                            | TokenError::IncompleteIdentifier
+                            | TokenError::IncompleteComment
```
jrvidal, line 134: :ringed_planet: Small unrelated fix: incomplete comments also count as multi-line input in the REPL.

Note that we cannot do that for unclosed escaped literals, b/c newlines are not accepted.

crates/steel-parser/src/lexer.rs
```diff
@@ -147,38 +150,47 @@ impl<'a> Lexer<'a> {
...
-                        }
-                        'a'..='f' | 'A'..='F' => {
-                            digits.push(c);
-                        }
-                        _ => return Err(TokenError::MalformedByteEscape),
+                        c if c == escape_end => break true,
+                        // note that this overlaps partially with `escapeEnd`
+                        ';' | '\\' | '\n' | '(' | ')' | '[' | ']' | '{' | '}' => break false,
+                        c if c == delim => break false,
+                        _ => digits.push(c),
```
jrvidal, line 175: The logic in this loop has changed significantly. Now we don't fail as early as we find an invalid character. Instead we try to continue as much as we can until we find some kind of "delimiter".

That way we can report unclosed literals. And in the case with invalid characters gets caught anyway by the `u32` parsing.

PR #501, opened 2025-09-07
improve port `write`

m4rch3n1ng:
previously in all calls to `write` (except when writing a single char), the buffer would be flushed. this essentially eliminates any performance benefit of even using a BufWriter, as now `write-string`, `write-byte`, `#%raw-display` etc are not actually buffered writes. additionally r7rs includes a procedure to explicitly flush the output port, so i don't think it should be flushed implicitly (and guile for example doesn't either). now it doesn't flush implicitly either.

doing this then vastly simplifies the write part of the ports file, as i can now remove the `write_without_flush` function, as the write function no longer flushes, and can remove the `write_and_flush!` macro.

additionally switch from using `write` to `write_all`. the [`write` docs](https://doc.rust-lang.org/beta/std/io/trait.Write.html#tymethod.write) mention that an error of kind `Interrupted` is not fatal and the write op should be retried, which is exactly what [`write_all` does](https://doc.rust-lang.org/beta/std/io/trait.Write.html#method.write_all), so just use that one.

### DIFF
--- a/crates/steel-core/src/values/port.rs
+++ b/crates/steel-core/src/values/port.rs
@@ -382,10 +382,8 @@ impl SteelPortRepr {
 
     pub fn write_char(&mut self, c: char) -> Result<()> {
         let mut buf = [0; 4];
-
         let s = c.encode_utf8(&mut buf);
-
-        let _ = self.write_without_flush(s.as_bytes())?;
+        let _ = self.write(s.as_bytes())?;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


mattwparas:
In general I'm fine with this change, however now callers who were expecting flushing to occur implicitly will need to do it explicitly - meaning, `display` and `displayln` will need a call to flush as well, or anything else here that is using these I believe. Could we add those in?

mattwparas (approved):

PR #505, opened 2025-09-10
misc port improvements

m4rch3n1ng:
i've been working on a bit of a refactor to the ports module for a new feature, where i wanted/needed these improvements, but i wanted that pr to be mostly on-topic of that feature, so here are just the slight adjustments so the next pr stacked on top of this will be more focused.

the improvements are these:

- [79276bd39715cd6f28be7a6bf317a68889558767]: the `read_byte` function had a return type indicating, that it would return a `MaybeBlocking`, but actually didn't check for the `WouldBlock` error kind, so fix that here.

- [e9f983a185ee2c581cc6a3bfb1bf352814eaadf6]: the `read_bytes_amt` fn returned a boolean, that i wasn't sure what it was doing and it was never used, so this pr removes that return parameter. there is quite a bit more wrong with this function, which i will fix in the upcoming pr after this one, as i need the refactor for those fixes, but when i wanted to actually fix that, i stumbled upon the boolean return parameter, where i wasn't sure what value i have to return in my case. then i went to investigate where it is used to figure out what i should do, and it turns out that it is not used at all.

- [312c8468e84dc0d29882737e40cf7a4cc81f75b6]: unrelated to the refactor, i added some tests i forgot to add in #492, that i wasn't sure where else i should put them and i didn't want a separate pr just for tests, so here they are.

### DIFF
--- a/crates/steel-core/src/values/port.rs
+++ b/crates/steel-core/src/values/port.rs
@@ -179,6 +179,30 @@ macro_rules! port_read_str_fn(
     }};
 );
 
+fn read_full<R: Read>(mut reader: R, mut buf: &mut [u8]) -> Result<MaybeBlocking<usize>> {
+    let mut amt = 0;
+    while !buf.is_empty() {
+        match reader.read(buf) {
+            Ok(0) => break,
+            Ok(n) => {
+                amt += n;
+                buf = &mut buf[n..];
+            }
+            Err(err) if err.kind() == io::ErrorKind::Interrupted => {}
+            Err(err) if err.kind() == io::ErrorKind::WouldBlock => {
+                if amt == 0 {
+                    return Ok(MaybeBlocking::WouldBlock);
+                } else {
+                    break;
+                }
+            }
+            Err(err) => return Err(err.into()),
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


mattwparas:
Can you elaborate on what is wrong with these functions?

m4rch3n1ng:
the `read` call is not great, as a `read` can just return an `Interrupted` ErrorKind, which should then retry the `read` op. i also cannot use `read_exact`, as that doesn't return the amount of bytes read and also leaves the buf in an "unspecified" state when it returns an error of kind `UnexpectedEof`. additionally, the call `read` should not return an error of kind `UnexpectedEof`, that is what the `read_exact` fn returns, the `read` fn should return an `Ok(0)` on eof.

for that i made a small function that's kind of like `read_exact`, but actually returns the amount read if it hits an eof.

now that i think about it, i could probably move the fix for the `read_bytes_amt` fn here if you prefer, so you can see what the other changes are for? give me a second though, it'll need quite a bit of a rebase.

m4rch3n1ng:
i was wrong, it was actually quite easy to cherry-pick that commit from the other wip branch and it was actually almost conflict-free.

PR #519, opened 2025-09-16
fix a bunch more clippy lints

m4rch3n1ng:
this fixes all clippy lints in the crates that are not steel-core and then hacks at two very low-hanging lints in steel-core.


mattwparas (approved):
thanks!

PR #531, opened 2025-09-24
make symbol=? work with uninterned symbols

m4rch3n1ng:
similar fix as #450, but for symbol=?.

the scheme spec says, that `eqv?` should return #t for two symbols, if
they are equal according to `symbol=?`. previously, `symbol=?` would
return `#f`, if the symbols were `string=?`, but one (or both) of the
symbols were uninterned. this was in conflict with `eqv?`, which would
return `#t` regardless of if the arguments were interned or not.

*but* the scheme spec does not say, that symbols must be `symbol=?` according to string equality, when they are interned. we've already had this discussion in #323, where we decided to just do ptr equality for `symbol=?` (and `eqv?`), because steel actually has symbol interning.

in racket, `(symbol=? 'a (string->symbol "a"))` actually returns `#t` even though racket only does ptr equality, as racket's `string->symbol` actually returns an interned symbol and you have to explicitly call `string->uninterned-symbol` if you want to get an uninterned symbol, unlike steel, where `string->symbol` returns an uninterned symbol. doing it like this should be doable in steel, if you want me to do that, but i wanted to get feedback on this first.

as a third option, i can also revert #450, but (as evident by someone opening an issue about it), the previous implementation was a little confusing, so i don't think that is a great option.

(additionally i was able to simplify the `eqv?` impl for symbol equality, as both the std Arc and the triomphe Arc already implement PartialEq by first checking for pointer equality and then for equality of the T, so we don't need to manually check for pointer equality ourselves.)


mattwparas:
So if you want `string->symbol` to return an interned symbol, you could take this function as inspiration:

```rust

#[steel_derive::context(name = "#%intern", arity = "Exact(1)")]
pub fn intern_symbol(ctx: &mut VmCore, args: &[SteelVal]) -> Option<Result<SteelVal>> {
    let mut guard = ctx.thread.compiler.write();
    let arg = args[0].clone();
    if let SteelVal::Void = arg {
        return Some(Ok(arg));
    }

    let interned_index = guard.constant_map.add_or_get(arg);
    let value = guard.constant_map.get(interned_index);
    Some(Ok(value))
}
```

We could do this for `string->symbol` and then otherwise make a `string->uninterned-symbol` like racket which just does what the current `string->symbol` does.

m4rch3n1ng:
made `string->symbol` return an interned symbol and created a new `string->uninterned-symbol` procedure.

i have also updated the pr description to reflect what the pr does now.

PR #532, opened 2025-09-25
validate macro templates

jrvidal:
Statically validates templates in syntax patterns to have the appropriate depth of ellipses/repetition.

This is not necessarily mandated by R7RS, but both Racket and Guile do validate this statically, and I think it just makes sense. This allows us to print some fancy error messages:
```
λ > (define-syntax foo (syntax-rules () [(_ (a ...) ...) (a ...)]))
error[E06]: BadSyntax
  ┌─ :1:55
  │
1 │ (define-syntax foo (syntax-rules () [(_ (a ...) ...) (a ...)]))
  │                                                       ^ missing ellipsis: pattern variable needs at least 2 levels of repetition, found 1
```

### DIFF
--- a/crates/steel-core/src/parser/macro_template.rs
+++ b/crates/steel-core/src/parser/macro_template.rs
@@ -1,0 +1,132 @@
+use std::ops::ControlFlow;
+
+use fxhash::FxHashMap;
+use steel_parser::{
+    ast::{Atom, ExprKind, List, Vector},
+    interner::InternedString,
+    parser::SyntaxObject,
+    tokens::TokenType,
+};
+
+use crate::{compiler::passes::VisitorMutControlFlow, SteelErr};
+
+pub struct MacroTemplate {
+    bindings: FxHashMap<InternedString, u8>,
+    depth: u8,
+    result: Result<(), SteelErr>,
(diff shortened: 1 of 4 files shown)


crates/steel-core/src/parser/macro_template.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn visit_vector(&mut self, v: &Vector) -> ControlFlow<()> {
+        if v.bytes {
+            return ControlFlow::Continue(());
+        }
+
+        self.visit_list_elements(&v.args, ListType::Vec)
+    }
+
+    #[inline]
+    fn visit_begin(&mut self, begin: &Begin) -> ControlFlow<()> {
```
jrvidal, line 120: Hmm, I'm thinking now that I should probably include other forms here... Ugh, how I would like to just deal with unlowered datums.

jrvidal: Yes. The many expansions phases in the compiler make my head hurt, but I sort of assume (as we've discussed before) that this is too early to have "high-level" AST nodes (e.g. `begin`, `let`, etc.).

PR #549, opened 2025-11-01
Implement LSP function to list symbols in current document

shybovycha:
Implement crude top-level symbol definition lookup in file - currently distinguishes top-level constants, top-level functions and `let` bindings.

For testing, I used [`flash.hx`](https://github.com/shybovycha/flash.hx/blob/master/flash.scm):

Screenshot 2025-11-02 at 7 47 58 pm

### DIFF
--- a/crates/steel-language-server/src/backend.rs
+++ b/crates/steel-language-server/src/backend.rs
@@ -17,12 +17,12 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use steel::{
     compiler::{
-        modules::{MANGLER_PREFIX, MODULE_PREFIX, steel_home},
+        modules::{steel_home, MANGLER_PREFIX, MODULE_PREFIX},
         passes::analysis::{
             query_top_level_define, query_top_level_define_on_condition,
             require_defitinion_to_original_symbol, Analysis, IdentifierStatus,
             RequiredIdentifierInformation, SemanticAnalysis, SemanticInformation,
-            SemanticInformationType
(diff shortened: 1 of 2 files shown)


shybovycha:
@mattwparas could you take a look at this one as well? i reckon with this one in the development experience with Steel could be improved significantly

crates/steel-language-server/src/backend.rs
```diff
@@ -258,6 +260,126 @@ impl LanguageServer for Backend {
...
+                                        _ => Some("function".to_string())
+                                    }
+                                },
+                                SemanticInformationType::CallSite(c) => Some("call".to_string()),
+                                SemanticInformationType::Let(l) => Some("let".to_string()),
+                            }
+                        });
+
+                        SymbolInformation {
+                            name: name.resolve().into(),
```
mattwparas: For the name, I think you'll want to actually get this from the rope itself, so take the span and index into the rope. Otherwise the name might be mangled - in the screen shot you sent the local variables are mangled. The original span should remain otherwise unchanged, so it should match what the identifier was originally

shybovycha: yup, great suggestion! thanks!
Screenshot 2025-11-04 at 4 27 43 pm
Screenshot 2025-11-04 at 4 31 22 pm

mattwparas (approved):

mattwparas:
Can you please run a `cargo fmt` on this? Otherwise LGTM

PR #551, opened 2025-11-03
Transducer implementation, find reference in LSP

mattwparas:


shybovycha:
@mattwparas i am trying the `references` implementation now and i can't get it to work - it just does not retrieve references for even the simple case:

Screenshot 2025-11-04 at 11 19 58 am

also does not seem to work for `let` bindings:

Screenshot 2025-11-04 at 11 19 45 am

from the logs i can see it fetches the *definition* under cursor, but not the *references*

shybovycha:
not much in the logs:

```
2025-11-04T11:58:28.986 helix_term::application [INFO] window/logMessage: LogMessageParams { typ: Info, message: "file opened!" }
2025-11-04T12:01:09.095 helix_view::editor [INFO] Switching view: DocumentId(2)
2025-11-04T12:01:18.300 helix_view::editor [INFO] Switching view: DocumentId(1)
2025-11-04T12:01:19.942 helix_lsp::transport [INFO] steel-language-server -> {"jsonrpc":"2.0","method":"textDocument/references","params":{"context":{"includeDeclaration":true},"position":{"character":23,"line":11},"textDocument":{"uri":"file:///Users/artem.shubovych/.config/helix/flash.scm"}},"id":1}
2025-11-04T12:01:19.948 helix_lsp::transport [ERROR] steel-language-server err <- "Unable to find a refers to for this identifier\n"
2025-11-04T12:01:19.950 helix_lsp::transport [INFO] steel-language-server <- {"jsonrpc":"2.0","result":null,"id":1}
2025-11-04T12:01:22.718 helix_view::editor [INFO] Switching view: DocumentId(2)
```

from what i can tell, backend fails to retrieve the information for _any_ symbol under cursor (based on where the error message is thrown from).

i will add more logs to look into it after lunch (in about an hour)

mattwparas:
Other useful information how are you opening the file? Is the working directory of helix a parent to the file?

shybovycha:
the way i have it set up is:

i have a fork of `steel`, which i use to install LSP with `cargo xtask install`

i have a separate fork of `hx` on your branch (`steel-event-system`), which i use to run Helix with `RUST_BACKTRACE=1 cargo run -- ~/.config/helix/flash.scm  -v`

helix fork, helix configs and steel fork all live in different directories, none of them are subdirectories of one another

shybovycha:
as you pointed out, i only see one `Finished indexing` message in my logs: 

> 2025-11-04T12:57:04.020 helix_lsp::transport [ERROR] steel-language-server err <- "Finished indexing workspace\n"

mattwparas:
Fix should be on master now - it was an accidental `||` instead of `&&` :smile:

shybovycha:
confirmed: works.

i have noticed one weirdness: sometimes the first call does not work, but second time (and onwards) it does - i presume it might be caused by the asynchronous nature of the code?

mattwparas:
No UI indication yet, but yeah, the work done progress is what will come next

Issue #535, opened 2025-10-05
Please review: initial `no_std` scaffolding across crates


mparkachov:
Hello Steel community! I’m teaching myself Rust and have been experimenting with Steel on an ESP board. The PoC ran with `std` because that board exposes a few standard APIs, but I quickly started hitting memory ceilings. To see if we can slim things down I sketched out what a `no_std` build might look like. This issue is an initial proposal to sanity check the approach and to hear whether full `no_std` support is something we want to pursue.

## Overview
- Introduces dedicated CI coverage for `no_std` targets with custom rustflags and a wasm test harness (`.github/workflows/no_std.yml:1`, `.cargo/no_std.toml:1`, `scripts/no_std_runner.js:1`).
- Reworks `steel-core`’s feature graph so `std` dependencies are optional and adds minimal fallbacks plus a custom `getrandom` shim to keep the crate compiling without `std` (`crates/steel-core/Cargo.toml:115`, `crates/steel-core/src/lib.rs:1`, `crates/steel-core/src/minimal.rs:1`, `crates/steel-core/src/getrandom_custom.rs:1`).
- Splits std-only functionality into gated modules for supporting crates and supplies placeholder implementations for `no_std` builds (`crates/cargo-steel-lib/src/lib.rs`, `crates/cargo-steel-lib/src/minimal.rs`, `crates/steel-gen/src/lib.rs`, `crates/steel-parser/src/lib.rs`, `crates/steel-parser/src/pretty.rs`).
- Adds a portable test harness so core smoke tests can run under `no_std` with custom allocator support (`crates/steel-core/tests/no_std_suite.rs:1`, `crates/steel-core/tests/portable.rs:1`).

## Notes
- Parser builds in the workflow still run under `continue-on-error`, so the `no_std` story there is incomplete (`.github/workflows/no_std.yml`).
- Some pieces are stubbed just to unblock compilation (e.g., the cargo-steel-lib runner currently returns an unsupported error in `no_std`, `crates/cargo-steel-lib/src/minimal.rs`).

This is an initial proposal whose goal is simply to get the workspace compiling with `no_std`. 

## Next steps
1. Split the regression suite so both `std` and `no_std` modes get dedicated coverage, using the wasm runner to surface real pass/fail signals.
2. Replace temporary stubs (cargo-steel-lib runner, parser pretty-printing, deterministic getrandom shim) with working implementations or tighter cfg gates.
[...]

PR #557, opened 2025-11-05
Enable no_std builds and expand platform abstractions across steel

mparkachov:
I put together changes which allow build of steel-core for no_std configuration as per #535. I assumed that alloc is still available. Tests were gated with std feature, so part of them now run in no_std build. I want to stress, that these tests still bring default test harness, which pulls std, so running tests are not 100% guarantee that they will work in pure no_std mode. This could be done with custom harness, but I didn't implemented it yet, wanted to get first feedback. There are a lot of changes, but I wanted to submit PR with green build for no_std. From functional perspective they could be summarised as following:

- Make std opt-in across the core crates so the workspace builds with --no-default-features, including the new #![no_std] entry points and default feature adjustments in steel-core, steel-parser, and steel-gen (crates/steel-core/src/lib.rs (line 1), crates/steel-core/Cargo.toml (line 127), crates/steel-parser/src/lib.rs (line 1), crates/steel-gen/src/lib.rs (line 2)).
- Add crate-local shims that replace std functionality when absent—covering collections, sync primitives, paths, OS strings, deterministic time, and randomness—to centralize environment differences (crates/steel-core/src/collections.rs (line 1), crates/steel-core/src/sync.rs (line 1), crates/steel-core/src/path.rs (line 1), crates/steel-core/src/os_strings.rs (line 1), crates/steel-core/src/time.rs (line 1), crates/steel-core/src/rand.rs (line 1)).
- Refactor compiler, VM, and primitive modules to lean on those shims, tighten cfg(feature = "std") guards around I/O and OS integrations, and wire the new abstractions through the runtime (crates/steel-core/src/primitives/fs.rs (line 1), crates/steel-core/src/steel_vm/engine.rs (line 1), crates/steel-language-server/src/backend.rs (line 19)).
- Expand optional dependency surface (e.g., hashbrown, getrandom) and align feature gating so std-only integrations are marked optional while sync now depends on std (crates/steel-core/Cargo.toml (line 40), crates/steel-core/Cargo.toml (line 129)).
- Refresh CI workflows to modern actions and explicitly build/test the workspace under no-std, wasm32, sync, and redox targets, ensuring coverage also exercises the --no-default-features configuration (.github/workflows/rust.yml (line 20)).

Now cargo ... --no-default-features builds without std assumed. Following are green:
```
cargo test --all
cargo test --no-default-features
cargo build -p steel-core --target wasm32-unknown-unknown 
cargo build -p steel-core --target wasm32-unknown-unknown --no-default-features
```
There are some issue with nix job in actions, but the reason is that current stable version of rust and version of rust in nix are different and they report differently fmt issues. Seeing recent update on old rust, maybe project should set on particular version, or create several builds for different rust versions. 

There is modules feature defined, but I could not figure out from implementation what it is supposed to do, but feels like something related.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -135,24 +135,24 @@ jobs:
           cd cogs/
           cargo run -- install.scm ../contrib/slack
 
-  redox-build:
-    name: redox build
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          profile: minimal
-          override: true
-      - name: Build x86_64-unknown-redox
-        run: |
-          sudo apt-get update -y && sudo apt-get install fuse3 libfuse3-dev pkg-config -y
-          cargo install redoxer --locked
-          redoxer build
-  
+  # redox-build:
+  #   name: redox build
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - name: Checkout repository
+  #       uses: actions/checkout@v2
+  #     - name: Install Rust toolchain
+  #       uses: actions-rs/toolchain@v1
+  #       with:
+  #         toolchain: stable
+  #         profile: minimal
+  #         override: true
+  #     - name: Build x86_64-unknown-redox
+  #       run: |
+  #         sudo apt-get update -y && sudo apt-get install fuse3 libfuse3-dev pkg-config -y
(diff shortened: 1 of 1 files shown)


mattwparas:
👋 Thanks for putting in the time on this, this is a big change set, so apologies but it will take me a while to go through it.

Could you expand on your use case here? I'd just like to understand what your eventual goal is and make sure that moving forward steel can support that without a large maintenance burden (i.e. making sure that keeping steel `no_std` compliant doesn't become too difficult)

I'd like to have steel be `no_std`, so this is something I'd like to have 😄

PR #562, opened 2025-11-08
implement `memf`, `assf` and `findf` procedures

m4rch3n1ng:
`memf` and `assf` are versions of `member` and `assoc` respectively, that take a procedure instead of a value to compare against. the `findf` procedure is similar to `memf`, but it returns the item itself instead of a list where the item is the car.

i needed this in my `helix.scm` file, where i have the following code:

```scheme
(define rust-analyzer
    (findf (λ (client) (string=? (lsp-client-name client) "rust-analyzer")) (get-active-lsp-clients)))
```

and i implemented `findf` myself (since it isn't hard), but i thought, maybe i could upstream it and maybe add some more procedures (i needed `memf` like a few months ago, forgot for what).

(the names are stolen from racket.)

### DIFF
--- a/crates/steel-core/src/scheme/stdlib.scm
+++ b/crates/steel-core/src/scheme/stdlib.scm
@@ -44,6 +44,7 @@
          assq
          assv
          assf
+         findf
          filter
          even-rec?
          odd-rec?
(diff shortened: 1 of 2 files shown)


mattwparas (approved):
Thank you!

PR #566, opened 2025-11-12
Docs: Fix typo documentation for `dropping`

afternoon:
Just noticed a minor mistake in the docs. Easy to fix!


m4rch3n1ng (changes requested):
this will need to be changed here:

https://github.com/mattwparas/steel/blob/8bc2c6aba8d8047415b33a1a3fe4244a74ede2b8/crates/steel-core/src/primitives/transducers.rs#L435

and then the docs regenerated via `cargo xtask docgen`

PR #579, opened 2025-12-04
(partially) fix Hash impl / is_hashable and make pairs hashable

m4rch3n1ng:
avoid crashes like

```
λ > (struct thing (val))
λ > (hashset (thing (into-sum)))

thread 'main' (80079) panicked at crates/steel-core/src/rvals.rs:2141:18:
not implemented: Attempted to hash unsupported value: #<reducer>
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

because is_hashable previously did not check if all values in a struct are hashable.

additionally simplifies the discriminant hashing, fixes hashing for void and makes it possible to hash pairs.

the same still applies to the vec / list hash impls, as is_hashable does not check if all those keys are hashable, because it would be a non-neglible performance impact. it might be worth it to make all values hashable and just fall back to pointer hashing for things where nothing else makes a lot of sense.

### DIFF
--- a/crates/steel-core/src/rvals.rs
+++ b/crates/steel-core/src/rvals.rs
@@ -2094,6 +2094,7 @@ impl SteelVal {
             BoolV(_) | IntV(_) | CharV(_) | VectorV(_) | StringV(_) | SymbolV(_) | HashMapV(_)
             | Closure(_) | ListV(_) | FuncV(_) => true,
             CustomStruct(s) => s.is_hashable(),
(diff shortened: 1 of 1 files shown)


mattwparas:
yeah, if you'd like to go ahead and do that it would be helpful. I don't see any reason why everything couldn't be hashable in some way

m4rch3n1ng:
sure, i'll open a second pr after this one is through.

PR #599, opened 2026-01-14
Fix spellcheck.scm example.

nebkor:
Previously, when trying to run the `spellcheck` example, right before the repl launched, the following error popped up:

```
Generating the bk tree
error[E03]: TypeMismatch
   ┌─ spellcheck.rkt:83:8
   │
83 │       (trim line)))
   │        ^^^^ trim: Cannot convert steel value: (eof) to steel string
```

This just updates the `get-next-word` function in the example to see if `line` is an eof object.

### DIFF
--- a/examples/scripts/spellcheck.scm
+++ b/examples/scripts/spellcheck.scm
@@ -78,7 +78,7 @@
 ;; otherwise returns the trimmed word
 (define (get-next-word!)
   (define line (read-line-from-port *corpus-port*))
-  (if (symbol? line)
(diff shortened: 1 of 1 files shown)


nebkor:
> Thank you!

Thank YOU!

I was looking around for an embeddable interpreter for a game in Rust I'm just starting, and when I saw a Rust-native Scheme, I knew I'd hit the jackpot :) Scheme was my first loved language, so I'm looking forward to writing more soon!

PR #606, opened 2026-01-29
no_std: refactor random primitives to use getrandom

mparkachov:
Centralize randomness in steel/random without rand, keeping std/no_std behavior intact via getrandom features.

Changes:
- Replace rand usage with getrandom::fill directly in random.rs.
- Keep unbiased range generation with rejection sampling.
- Remove rand dependency; wire getrandom with default-features = false and getrandom/std under the std feature.

Testing:
- cargo test --all -> ok


mattwparas (approved):

mattwparas:
whoops - there are some merge conflicts

Otherwise this LGTM

mparkachov:
no progress

PR #608, opened 2026-01-29
no_std: add PathBuf abstraction

mparkachov:
Summary:
- Add steel::path::PathBuf wrapper and expose it from steel-core, plus a std-only path separator constant and component collection support in path.rs.
- Replace std::path::PathBuf usage across steel-core and dependent crates with steel::path::PathBuf, updating public APIs and call sites accordingly (compiler/modules, VM, parser, fs/http, docs, LSP, REPL, CLI).
- Make env_home, glob, xdg, and which optional deps gated by the std feature in Cargo.toml.

Testing:
- cargo test --all -> ok


crates/steel-core/src/steel_vm/engine.rs
```diff
@@ -59,7 +60,7 @@ use std::{
     borrow::Cow,
     cell::{Cell, RefCell},
     collections::{HashMap, HashSet},
-    path::{Path, PathBuf},
```
mattwparas: So my only concern with this change is that there are a bunch of public APIs that expect a `std` PathBuf, so this change would be breaking. In particular in the engine API. I suppose one fix would be to have the `std` feature just be a type alias for the standard PathBuf as to not break any callers?

mparkachov:
no progress

PR #613, opened 2026-02-05
fix: eliminated build warnings

mparkachov:
IMPORTANT: fixes are done mechanically

- Gated JIT-only helpers and imports to the jit2 feature and fixed JIT function-pointer casts in lists.rs, vm.rs and cgen.rs.
- Cleaned up unused types/imports and added meaningful Custom::fmt/Display impls to silence dead-code warnings in lib.rs.
- Removed the unused import in spellcheck.rs.

Tests: cargo test --all now compiles without warnings, tests are ok.


mattwparas:
Could you elaborate on "fixes are done mechanically"?

PR #650, opened 2026-03-17
add aliases to bytevector procedures for r7rs compat

m4rch3n1ng:
and also add an alias for `bytes-copy` for `bytevector-copy` for consistency.

additionally implement `bytes-read!` (with a `bytevector-read!` alias) and actually respect the parameter for `read-bytes-into-buf` by just implenting that via the `bytes-read!` procedure.


mattwparas (approved):

mattwparas:
thank you! sorry for the delay

PR #660, opened 2026-05-05
feat(fs): add `rename!` function

melindachang:
Added a function `rename!` to move files/directories, plus docs.

### DIFF
--- a/crates/steel-core/src/primitives/fs.rs
+++ b/crates/steel-core/src/primitives/fs.rs
@@ -86,6 +86,7 @@ pub fn fs_module() -> BuiltInModule {
     module
         .register_native_fn_definition(DELETE_DIRECTORY_DEFINITION)
         .register_native_fn_definition(CREATE_DIRECTORY_DEFINITION)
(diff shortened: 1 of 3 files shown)


crates/steel-core/src/primitives/fs.rs
```diff
@@ -369,6 +370,26 @@ pub fn create_directory(directory: &SteelString) -> Result<SteelVal> {
...
+/// * source : (string?) - The file or directory to rename.
+/// * destination : (string?) - The destination to which to move the file or directory.
+///
+/// # Examples
+/// ```scheme
+/// > (rename! "logs/today.json" "logs/tomorrow.json") ;;
+/// > (rename! "logs" "backup") ;;
+/// ```
+///
+#[steel_derive::function(name = "rename!")]
```
mattwparas: could we use `rename-file-or-directory`? That way its clear that its applying to a file or directory (if a little verbose)

melindachang: Yep, fixed

mattwparas (approved):

PR #677, opened 2026-07-23
Make serialized programs runnable in a fresh engine (constant map merge + faithful round trip)

koalazub:
Found these while building startup caching for the [nothelix plugin](https://github.com/koalazub/nothelix) on the [helix fork](https://github.com/koalazub/helix) that I use. We compile the whole plugin as one RawProgramWithSymbols, serialise it to disk, and run it from the artifact on later launches so the editor doesn't recompile ~9k lines of scheme every start at this pint of writing.

Two things came up on that path.

Running a deserialised program installs its constant map as the VM's live table while Compiler::constant_map stays separate. The program itself runs fine, but any eval afterwards compiles constant indices against the compiler's map and resolves them in the program's map, so you get FreeIdentifier or the wrong literal back. The fix rebases a detached program onto the compiler's live map in raw_program_to_executable (add_or_get and rewriting the constant referencing payloads), and drops the map swap in execute_non_interactive_program_image since it's covered by the same path now.

ConstantMap's to_bytes/from_bytes went through display strings and a reparse, which re-lowers on the way back in. (quote x) came back as x and let forms came back as lambdas, so a round trip changed the program.

note: merge_constants_into has to know which opcodes reference constants (PUSHCONST, EQUALCONST and the register PASS variants). I checked the list against the current vm.rs but it's the kind of thing that could rot quietly if new opcodes land. Happy to restructure if you'd rather that knowledge live somewhere else.

### DIFF
--- a/crates/steel-core/src/compiler/compiler.rs
+++ b/crates/steel-core/src/compiler/compiler.rs
@@ -577,7 +577,7 @@ impl Compiler {
     pub(crate) fn into_serializable_compiler(self) -> Result<SerializableCompiler> {
         Ok(SerializableCompiler {
             symbol_map: self.symbol_map,
-            constant_map: self.constant_map.into_serializable_map(),
+            constant_map: self.constant_map.into_serializable_map()?,
             macro_env: self.macro_env,
             opt_level: self.opt_level,
             module_manager: self.module_manager,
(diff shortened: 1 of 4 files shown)


mattwparas:
Unfortunately this will still fail in very subtle ways if you do any more evaluation or interact with the kernel (certain kinds of macro expansion) since the state of the program is not correctly sync'd up with the rest of the engine. It will require quite a bit more work to properly handle all of the state along the way, alongside the module cache which isn't handled properly.

Issue #682, opened 2026-08-03
Library rust function with isize argument does not take values below 0


scrythe:
If I make a rust library and declare a function that takes isize as an argument, I can't actually execute the function with values below 0.

Below is code taken from crates/steel-core/src/steel_vm/ffi.rs

```rust
impl<'a> FromFFIArg<'a> for usize {
    fn from_ffi_arg(val: FFIArg<'a>) -> RResult<Self, RBoxError> {
        if let FFIArg::IntV(i) = val {
            if i < 0 {
                conversion_error!(usize, val)
            } else {
                RResult::ROk(i as usize)
            }
        } else {
            conversion_error!(usize, val)
        }
    }
}

impl<'a> FromFFIArg<'a> for isize {
    fn from_ffi_arg(val: FFIArg<'a>) -> RResult<Self, RBoxError> {
        if let FFIArg::IntV(i) = val {
            if i < 0 {
                conversion_error!(isize, val)
            } else {
                RResult::ROk(i as isize)
            }
        } else {
            conversion_error!(isize, val)
        }
    }
}
```

The usize one gives a conversion_error when the value is below 0 but that's to be expected. However the same is true for isize, giving a conversion_error when the value is below 0.

PR #683, opened 2026-08-03
allow negative values in isize argument conversion

scrythe:
When making a rust function for the steel library, a function that acceps an isize argument only allows positive values because the 
```rust
impl<'a> FromFFIArg<'a> for isize 
```
returns an error if the value is below 0, identical to how usize argument conversion works.
This pull request fixes that, allowing isize arguments to be below 0.


scrythe:
This fixes fix: #682

mattwparas (approved):

mattwparas:
Thank you!

PR #686, opened 2026-08-14
forge: drop args correctly on windows

ongyx:
This PR fixes a small bug where args are not dropped when running forge as a standalone script on Windows with `steel forge.scm`, as the interpreter has the .exe extension.


mattwparas (approved):

mattwparas:
Thank you!

PR #687, opened 2026-08-19
docs: fix doctest/example for Engine.register_type

waynr:
I noticed what looks like a copy/paste issue in the example/doctest for `Engine.register_type` where it looks like the example was at some point actually written for `Engine.register_fn`.

I've verified the doctest passes with `cargo test --docs -p steel-core steel_vm::engine::Engine::register_type`.


waynr:
It looks like the test suite ended up hanging and being cancelled, but I can't see the workflow logs to diagnose what, if anything, might be wrong with my PR.

mattwparas:
I believe its unrelated

